PR #1376, opened 2026-08-14
Replace execSync with argv-based spawnSync in codesign hook

anupamme:
## Summary

- Replaces shell-interpolated `execSync` calls in the `afterPack` codesign hook with `spawnSync` using argument arrays — eliminates shell interpretation entirely
- Adds a `runOrThrow` wrapper that throws on spawn errors or non-zero exit status, preserving `execSync`'s build-halting behavior
- No path sanitization or rewriting — `appOutDir` and `productName` are build-time values from electron-builder, not user input

This is a defensive cleanup, not a security fix. Semgrep flags the use of `child_process` in this hook, but the arguments are build-time values rather than end-user input. The change is still worthwhile: argv-based invocation is more robust against paths containing shell metacharacters, and the explicit error propagation makes failure modes clearer.

## Test plan

- [x] `npx vitest run scripts/codesign-ad-hoc.test.js` — 10 tests pass
- [ ] Verify macOS build still signs correctly (`npm run build` on mac)

### DIFF
--- a/scripts/codesign-ad-hoc.cjs
+++ b/scripts/codesign-ad-hoc.cjs
@@ -11,8 +11,18 @@
 // the xattr/detritus cleanup still runs (it's harmless and prevents codesign
 // failures on CI), but the ad-hoc --sign - step is skipped.
 
-const { execSync } = require('child_process');
+const { spawnSync } = require('child_process');
 const path = require('path');
+
+exports._spawnSync = spawnSync;
+
+function runOrThrow(command, args) {
+  const result = exports._spawnSync(command, args, { stdio: 'inherit' });
+  if (result.error) throw result.error;
+  if (result.status !== 0) {
+    throw new Error(`[codesign-ad-hoc] ${command} exited with status ${result.status}`);
+  }
+}
 
 exports.default = async function codesignAdHoc({ appOutDir, packager, electronPlatformName }) {
   if (packager.platform.name !== 'mac') return;
@@ -26,9 +36,9 @@ exports.default = async function codesignAdHoc({ appOutDir, packager, electronPl
   // xattr -cr handles all file types including symlinks inside Electron frameworks
   // (e.g. Versions/Current) that the previous find -not -type l approach skipped.
   console.log(`[codesign-ad-hoc] Removing ._* resource fork files from ${appPath}`);
-  execSync(`find ${JSON.stringify(appPath)} -name "._*" -delete`);
+  runOrThrow('find', [appPath, '-name', '._*', '-delete']);
   console.log(`[codesign-ad-hoc] Clearing xattrs on ${appPath}`);
-  execSync(`xattr -cr ${JSON.stringify(appPath)}`);
+  runOrThrow('xattr', ['-cr', appPath]);
   // MAS builds are always signed by electron-builder with the Mac App Store
   // distribution identity (from the keychain) — never ad-hoc sign them, or the
   // App Store signature would be invalid. The xattr cleanup above still helps.
@@ -41,5 +51,7 @@ exports.default = async function codesignAdHoc({ appOutDir, packager, electronPl
     return;
   }
   console.log(`[codesign-ad-hoc] Signing ${appPath}`);
-  execSync(`codesign --sign - --deep --force ${JSON.stringify(appPath)}`);
+  runOrThrow('codesign', ['--sign', '-', '--deep', '--force', appPath]);
 };
+
+exports.runOrThrow = runOrThrow;
--- a/scripts/codesign-ad-hoc.test.js
+++ b/scripts/codesign-ad-hoc.test.js
@@ -1,0 +1,121 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import path from 'path';
+import { createRequire } from 'module';
+
+const require = createRequire(import.meta.url);
+const mod = require('./codesign-ad-hoc.cjs');
+const codesignAdHoc = mod.default;
+const { runOrThrow } = mod;
+
+const mockSpawnSync = vi.fn();
+
+beforeEach(() => {
+  mockSpawnSync.mockReset();
+  mockSpawnSync.mockReturnValue({ status: 0, error: null });
+  mod._spawnSync = mockSpawnSync;
+  delete process.env.CSC_LINK;
+});
+
+describe('runOrThrow', () => {
+  it('does not throw when command succeeds', () => {
+    mockSpawnSync.mockReturnValue({ status: 0, error: null });
+    expect(() => runOrThrow('codesign', ['--sign', '-'])).not.toThrow();
+  });
+
+  it('throws when command exits with non-zero status', () => {
+    mockSpawnSync.mockReturnValue({ status: 1, error: null });
+    expect(() => runOrThrow('codesign', ['--sign', '-']))
+      .toThrow('[codesign-ad-hoc] codesign exited with status 1');
+  });
+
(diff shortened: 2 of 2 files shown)


krelltunez:
Thanks for the submission, but I'm closing this — the finding is a false positive, and the patch is a net regression.

**There is no untrusted input on this path.** `scripts/codesign-ad-hoc.cjs` is an electron-builder `afterPack` hook (wired at `electron-builder.config.cjs:55`). It runs at build time, on CI or a maintainer's machine. `appOutDir` and `packager.appInfo.productName` are supplied by electron-builder from this repo's own build config — they are never attacker-controlled. The PR's "Threat Model Context: This is a web application - XSS and injection vulnerabilities can affect end users" is boilerplate that does not apply to a build script; nothing here is reachable by an end user. The existing `JSON.stringify()` quoting was already adequate for the values that actually flow through.

Setting the false positive aside, the patch introduces real problems.

**1. It converts hard build failures into silent ones.** `execSync` throws on a non-zero exit; `spawnSync` does not — it returns `{ status, error }`, and this patch ignores both:

```
spawnSync('false')           → status 1,      no throw
spawnSync('missing-binary')  → error ENOENT,  no throw
execSync('false')            → throws
```

So if `codesign --sign -` fails, or `xattr`/`codesign` aren't on PATH, the build now continues and ships an unsigned `.app`. That is precisely the "damaged and can't be opened" Gatekeeper failure this hook exists to prevent — except it would now ship silently instead of failing the build. Moving to an argv-array `spawnSync` is a defensible idea on its own, but it requires checking `status`/`error` and throwing.

**2. The path-escape guard is unreachable.** `safeName` strips all `/` and `\`, so `path.join(resolvedOutDir, safeName + '.app')` is always a direct child of `resolvedOutDir`. The `if (!appPath.startsWith(resolvedOutDir + path.sep)) throw` can never fire — verified against `../evil`, `..`, `""`, `a/b`, and `a\b`. It's dead code.

**3. `appOutDir.replace(/\.\./g, '')` corrupts valid paths.** A legitimate output directory such as `build..v2/mac` silently becomes `buildv2/mac`, and the build then operates on the wrong path. Stripping a substring is the standard broken-sanitizer pattern, and it's redundant here — `path.resolve` plus a prefix check is the actual control. Same for `productName.replace(/[/\\]/g, '')`, which silently retargets rather than rejecting. Both contradict the PR's own "leaves valid inputs unaffected" claim.

**4. The "regression test" was never added, and could not pass if it were.** The diff is 1 file, +9/−5 — there is no test file. As written it cannot work:

- `import { codesignAdHoc } from '../scripts/codesign-ad-hoc.cjs'` — the module exports `exports.default`; there is no named `codesignAdHoc` export.
- Every payload passes `packager: {}`, so the first line, `packager.platform.name`, throws `TypeError` before any command is constructed. Every case lands in the `catch`.
- The catch asserts `expect(['string','object']).toContain(typeof error)`, which is true for any thrown value. The test passes vacuously against the *unpatched* file — it validates nothing.
- `global.child_process = {...}` intercepts nothing; the module captured the function via `require` at load time.
- It is jest-flavored with no imports; this repo runs vitest.

This is the second submission from this tooling here, after #1328 — which claimed CRITICAL / CWE-79 / "Likely exploitable" API key disclosure against `public/locales/en/translation.json`, a translations file containing no keys and no executable code. That was closed too.

I'd ask you to stop opening automated PRs against this repository. You engaged seriously with the feedback on #1328, which I appreciated — so this isn't about tone. It's that both submissions share the same two structural problems: a threat model that doesn't match what the code actually does, and a regression test in the description that isn't in the diff and wouldn't pass if it were. Until that changes, triaging these costs me more time than they save. If the tooling gets to the point where it's surfacing real issues in this codebase, I'm glad to look again.

anupamme:
Thanks for the detailed feedback. I agree that I overstated the security impact here; this hook runs in the Electron build environment, so the web-app/XSS threat model in my description isn’t applicable.

I also agree that my spawnSync change wasn’t preserving the existing failure semantics. I should have checked both `result.error` and the exit status so a failed codesign/xattr/find still fails the build.

The part I’d still like to pursue is replacing shell-string construction with argv-based execution. My intent there was to avoid relying on shell parsing for paths, rather than to claim that appOutDir is attacker-controlled.

If you're open, I can revise the PR to:

* preserve the existing error behaviour for non-zero exit codes;
* avoid mutating paths with the replace() sanitization;
* add proper Vitest coverage for spaces/special characters in paths and command failures;
* update the description to frame this as build-script hardening rather than an end-user vulnerability.

krelltunez:
That all sounds right, and I appreciate you separating the argv change from the threat-model claim. You're correct that it stands on its own. I checked, and `JSON.stringify` quoting doesn't actually protect against `$` or backticks inside shell double quotes, so argv is the more robust pattern regardless of whether anything here is attacker-controlled:

```
"dist/mac $HOME/App.app"  => shell saw: dist/mac /root/App.app
"dist/`id`/App.app"       => shell saw: dist/uid=0(root) gid=0(root)…
```

Nothing in the current config triggers that (`productName: 'dayGLANCE'`, `output: 'dist-app'`), so it's latent rather than a live bug, but it's a fair cleanup.

I've reopened this. Happy to take your first, second and fourth points as you described them.

On tests: I'd skip the spaces/special-character cases. Once you're passing an argv array there's no shell parsing left, so those would mostly assert that you called `spawnSync` with an array. The one with real value is failure propagation: that a non-zero exit from `codesign` or `xattr` actually fails the build. If that needs more mocking scaffolding than it's worth for a mac-only build hook, no test is fine; I'd rather have the six-line change clean than a test fighting the module boundary.

And to be clear about my earlier note, that was aimed at the automated pipeline output, not at you. A hand-checked revision like this is welcome.

anupamme:
Thanks for the detailed review. You're right on all counts. I overstated the threat model (these are build-time values, not user input), the path sanitisation could corrupt legitimate paths, and the spawnSync conversion silently swallowed failures.

I've reworked the PR as a smaller defensive cleanup:

  - Kept argv-based spawnSync (avoids shell interpretation of paths with metacharacters)
  - Added runOrThrow that throws on spawn errors or non-zero exit — preserves execSync's build-halting behaviour
  - Removed all path sanitization/rewriting
  - Added Vitest tests covering command success, failure, ENOENT, and the MAS/CSC_LINK skip paths

krelltunez:
This is exactly right. Thanks for reworking it.

I ran the tests rather than taking the description on faith, and they hold up:

- Mutating `runOrThrow` to swallow both `result.error` and the non-zero status check fails precisely the three failure-propagation tests. That was the regression in the first version, and it's now genuinely guarded.
- Reintroducing the `/\.\./g` strip fails the `..` test. I'd assumed that one was tautological, since both sides call `path.join`, but the mangled path resolves to a different directory, so it earns its place.
- Full suite green with the new file included: 78 files, 1144 tests.

The `exports._spawnSync` seam is slightly unusual, but `vi.mock` against CJS + `createRequire` would be worse, and it's two lines. No objection.

One thing before I merge. The second box in your test plan is still unchecked, and it's the one the unit tests can't reach. Everything mocks `spawnSync`, so the real `find`/`xattr`/`codesign` invocations are unexercised, and since this is mac-only, CI won't cover them either. I'll run a local mac build and merge once it comes back clean.

Also worth noting for the record: `stdio: 'inherit'` means these commands now write to the build log, where `execSync` previously captured and discarded stdout. Fine by me, since that's more visibility rather than less.

I really appreciate the way you handled the review.

PR #1406, opened 2026-08-17
Fix project membership for recurring tasks

### LINKED ISSUE #1405: [ENHANCEMENT]: Project membership lost on recurring tasks
### Reproduction steps:
1. Create a new task, assign it to a project
2. Set a recurrence and save

#### Expected:
The task stays in the project.
#### Actual:
It disappears from the project

The recurring template is created without `projectId` (the recurrence branch in `addTask` builds the template with an explicit field list that omits it). The field is also dropped when converting an existing task to/from recurring in the edit modal, and recurring occurrences vanish from project-filtered timeline/week views because expanded instances don't inherit `projectId` from the template.

I've created a fix and will make a PR. I also discovered that the delete dialogue for recurring tasks is drawn (`z-50`) below the Goal / Project view so fixed that in the PR as well.

vmpyr:
Closes: #1405

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -3552,6 +3552,7 @@ const DayPlanner = () => {
             notes: template?.notes || '',
             subtasks: template?.subtasks ? JSON.parse(JSON.stringify(template.subtasks)) : [],
             date: newTask.date || parsed.dateStr,
+            projectId: newTask.projectId || undefined,
             ...(mobileEditingTask.assignedUserSyncIds?.length ? { assignedUserSyncIds: mobileEditingTask.assignedUserSyncIds } : {}),
           };
           setTasks(prev => [...prev, regularTask]);
@@ -3585,6 +3586,7 @@ const DayPlanner = () => {
               completed: false,
               notes: '',
               subtasks: [],
+              projectId: newTask.projectId || undefined,
               assignedUserSyncIds: mobileEditingTask.assignedUserSyncIds,
             }]);
           } else {
@@ -3629,6 +3631,9 @@ const DayPlanner = () => {
                 ...t,
                 exceptions,
                 assignedUserSyncIds: templateAssigned,
+                // Project changes in the editor apply to the whole series —
+                // project membership is series-level, like the template itself.
+                projectId: newTask.projectId || undefined,
                 // Update recurrence pattern on template if changed
                 recurrence: { ...newTask.recurrence, startDate: t.recurrence?.startDate || parsed.dateStr.substring(0, 8) + '01' },
                 lastModified: new Date().toISOString(),
@@ -3653,6 +3658,7 @@ const DayPlanner = () => {
         recurrence: { ...newTask.recurrence, startDate: taskDate },
         completedDates: existingTask?.completed ? [taskDate] : [],
         exceptions: {},
+        ...(newTask.projectId ? { projectId: newTask.projectId } : {}),
         assignedUserSyncIds: mobileEditingTask.assignedUserSyncIds ?? existingTask?.assignedUserSyncIds,
         lastModified: new Date().toISOString()
       };
@@ -5982,6 +5988,9 @@ const DayPlanner = () => {
           isRecurring: true,
           recurringTemplateId: template.id,
           recurrenceType: template.recurrence?.type,
+          // Project membership is series-level (stored on the template);
+          // instances inherit it so project-filtered views keep occurrences.
+          projectId: template.projectId,
           ...(template.isExample ? { isExample: true } : {}),
         });
       }
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -3643,8 +3643,11 @@ const DayPlanner = () => {
         }
       }
     } else if (newTask.recurrence) {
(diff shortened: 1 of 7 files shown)


krelltunez:
Thanks for this. You found the bug, diagnosed it correctly, and fixed a second one on the way. Much appreciated!

I verified the branch locally: lint is clean across all seven changed files (`eslint --max-warnings 0`), and merged into current `main` the suite is green at 1939/1939. It merges cleanly, and none of the files you touched have drifted since your base. CI is green now too. The earlier red `check` was a stale merge base on our side, nothing to do with your changes.

A couple of things I want to call out as right, because they're the parts a partial fix would have gotten wrong:

- The expansion change and the editor write-back are correctly coupled. `openMobileEditTask` seeds `newTask.projectId` from `task.projectId`, so `projectId: newTask.projectId || undefined` in the edit branch is only safe *because* instances now carry `projectId`. Ship one without the other and every save silently strips the project. Good that both are here.
- The ghost-task fix is a genuine second bug. `ProjectCard` and `ProjectPlanner` open unscheduled project tasks with `isInbox=false`, so the recurrence branch ran and only filtered `tasks`.
- The z-index change is justified rather than cargo-culted. `z-50` sat under `GoalDashboard` (`z-[60]`), `ProjectPlanner` (`z-[70]`) and the task editor (`z-[80]`).

And to answer a question you didn't ask but might be wondering about: `RecurringSeriesRow` is welcome, not scope creep. Without it the fix is invisible where it matters most, since the project card enumerates only `tasks` + `unscheduledTasks`, so a project whose work is a weekly series would still look empty. Please keep it.

## Requested changes

All small. Should be about four lines plus two one-liners.

**1. i18n, with no translation work needed. The strings already exist.**

Heads up on a trap here: `src/locales.test.js` enforces strict two-way key parity across all seven locales, so adding `planner.recurring` to `en` alone would turn CI red. You don't need to. All three strings are already in the bundles, translated in every language:

- `sched.recurring` gives Recurring / Wiederkehrend / Recurrentes / Récurrentes / Ricorrenti / Recorrentes
- `sched.markComplete` and `sched.markIncomplete`

So in `RecurringSeriesRow.jsx` and `ProjectPlanner.jsx`:

```diff
-  {t('planner.recurring', 'Recurring')}
+  {t('sched.recurring')}
```

```diff
-  aria-label={completed ? 'Mark incomplete' : 'Mark complete'}
+  aria-label={completed ? t('sched.markIncomplete') : t('sched.markComplete')}
```

(`RecurringSeriesRow` will need `useTranslation` wired in.)

**2. One expansion site missed.**

`overdueRecurringAllDay` in `App.jsx` (around line 2968 on current `main`) is a fourth explicit field list building recurring instances, and it still omits `projectId`. Same class as the bug you're fixing. The visible effect: an overdue all-day recurring task in a project renders with no project pill in SCHED, since `SchedTaskCard` is passed `showProject` there. Same one-liner as your `expandedRecurringTasks` hunk.

I checked the other expansion sites (`useStats`, `useConflictDetection`, `useTaskDerived`, `WeeklyReviewModal`, `trmnl`) and those are counting and conflict-geometry, so they genuinely don't need it. Don't worry about those.

**3. Hide ended series.**

`RecurringSeriesRow` currently renders any template whose `projectId` matches, including one past its `endDate` or with `maxOccurrences` exhausted, so a finished series becomes permanent clutter in the card. Gating on `getNextOccurrence(template) !== null` (already exported from `recurrenceEngine.js`) should cover it.

**4. Nit: duplicate `Repeat` icon.**

It renders as the left-hand indicator when the series doesn't occur today, and again as a right-side badge when it does. One or the other.

## Not your problem. We'll take these as follow-ups

Listing them so you know they're deliberate omissions rather than things I missed:

- `projectId` in `src/utils/mcpReadModel.js`, which touches the MCP read-model contract and has its own spec docs. Ours to own.
- The removal from `unscheduledTasks` doesn't record a tombstone, so with cloud sync another device could resurrect the original next to the new series. The pre-existing `setTasks` removal has the same gap, so this isn't a regression you introduced.
- Excluding recurring from `calculateProjectProgress` is the right call, since a series never completes, but it means a project made only of series shows a 0% bar, and `isProjectStalled` can false-positive on a mixed project kept alive purely by a recurring task. Both ours.

Happy to take any of the four requested items myself if you'd rather not. Just say so and I'll push them to the branch. Otherwise this looks good to merge once they're in.

krelltunez (changes requested):
Requesting changes to mark the four items in my comment above as blocking. Nothing structural, all small, and the fix itself is right. Happy to merge once they're in.

vmpyr:
sure, I'll push the requested changes :)

krelltunez (approved):
All four addressed, thanks. Verified locally against current `main`: lint clean, 1939/1939 green. Merging.

Two small things I'll pick up separately, both consequences of the ended-series filter I asked for rather than anything you did: the filter is unmemoized in `ProjectCard` (which re-renders every minute), and `getNextOccurrence` builds a two-year range to return one date. I'd rather fix that in `recurrenceEngine` than have you work around it here.

PR #1604, opened 2026-09-11
Add selective Todoist task sync

Lcub3d:
## Summary

Adds an optional Todoist integration for selective task import into dayGLANCE, with guarded completion writeback. Todoist supplies tasks; dayGLANCE remains the place to schedule them.

## Scope

- Three modes: Today's tasks, All active tasks, and Choose by project, label or priority.
- Advanced filters support projects/subprojects, labels, priorities and AND/OR between selected groups; empty advanced filters import nothing. Overdue tasks are opt-in.
- Imported tasks can follow their Todoist date, go to Today, or remain in the inbox. Stable account-qualified IDs prevent duplicate imports and manually arranged time blocks are preserved.
- Connecting loads a preview; importing requires Sync now or explicitly enabled automatic sync. Automatic sync only runs while the app is open, visible and online.
- Desktop settings follow the existing persisted collapse convention; the mobile settings page remains independent.
- Mirror mode is deferred, as requested. This version never automatically deletes or restores local tasks because of filter changes or remote deletion. Old experimental settings migrate with automatic sync and writeback disabled.

## Safety and privacy

- Opt-in and disabled in multi-user mode. No third-party proxy and no upload of native dayGLANCE tasks.
- Credentials are session-only; the account-specific sync cache is in localStorage. Imported copies may participate in normal dayGLANCE backups and sync.
- Before cache persistence, prune unlinked completed/deleted tasks and deleted projects/labels; retain inactive task records needed by local copies or pending commands. The cache still shares the browser's storage budget.
- Disconnect cancels outstanding work, disables automation/writeback, removes session credentials, and clears cache under the writer lock when no pending receipts remain. Pending command UUIDs are retained for safe retry; imported local tasks are kept.
- Missing source records mean no information, not deletion. Completion writeback only closes eligible non-recurring leaf tasks, with durable command IDs and Web Locks; it does not write titles, dates, deletions or undo-completion operations.

## Review follow-up

- Extracted pure applyPlannedList into src/utils/todoistReconciliation.js with direct concurrent-edit tests.
- Added retention/disconnect tests, standard settings dividers, cancellation-specific errors, and docs/todoist-integration.md; expanded dense reconciliation/reason logic.
- Storage breakdown recognizes account-specific Todoist cache keys. Its storage.todoistCache label now lives in all eight existing lazy locale bundles, fixing the translation-key CI failure without weakening the checker. Eight rendering tests exercise those real bundles without registering the inline Todoist namespace.

## Validation

- Latest upstream [CI run](https://github.com/krelltunez/dayGLANCE/actions/runs/34683978461): dependency audit, lint, full tests and production build passed.
- Local focused Todoist/reconciliation/storage/locales suite: 193/193 passed; lint and web, Android WebView and Electron renderer builds passed.
- The unfiltered local Windows suite was not all green: 3,505 passed and 37 failed in Electron path, solar/timezone and Obsidian scenario tests. No matched clean-main Windows baseline was run, so these are not claimed as proven baseline failures. The upstream Linux merge CI is green.
- No live Todoist account end-to-end test or native installer build was performed in this follow-up.

## Translation handoff

The existing English/zh-CN Todoist namespace remains for the maintainer's planned migration after merge. Only the shared storage-breakdown label was moved now to fix its standalone loading and CI. Chinese quick-add vocabulary and Plan-vs-Do/Jobo ideas are separate follow-ups, not additions to this PR.


src/hooks/useTodoistSync.js
```diff
@@ -0,0 +1,237 @@
...
+    const enteringMirror = patch.mode === 'mirror' && latest.current.settings.mode !== 'mirror';
+    const changingMirror = (patch.mode || latest.current.settings.mode) === 'mirror' &&
+      ['mirrorScope', 'mirrorDates', 'priorities', 'projects', 'labels', 'match', 'labelMatch', 'subprojects', 'todayOnly', 'includeOverdue']
+        .some(field => Object.hasOwn(patch, field));
+    const next = normalizeSettings({ ...latest.current.settings, ...patch,
+      ...((enteringMirror || changingMirror) ? { mirrorAcknowledged: false, completionWriteback: false } : {}) });
+    try { writeJSON(localStorage, CONFIG_KEY, next); latest.current.settings = next; setSettings(next); setError(''); }
+    catch (err) { setError(err.message); }
+  }, [cancel]);
+  const disconnect = useCallback(() => {
```
krelltunez, line 49: blocking: Disconnect leaves the cache on disk.

This clears the token and account from sessionStorage, but never touches `stateKey(id)` in localStorage. A full copy of the user's Todoist account then persists indefinitely, with no way to remove it short of a full app reset.

I understand wanting to keep pending write receipts, and the UI does say so. But for a feature whose pitch is privacy there should be some way to purge. Either clear the cache when the outbox queue is empty, or add an explicit "forget cached Todoist data" action alongside disconnect.

src/hooks/useTodoistSync.js
```diff
@@ -0,0 +1,237 @@
...
+  const blockedWrites = catalog ? local.filter(task => ['recurring', 'parent', 'assignedElsewhere'].includes(writebackReason(task, catalog, settings))) : [];
+  const plannedRemovals = catalog ? local.filter(task => mirrorRemoval(task, catalog, { ...settings, mirrorAcknowledged: true }, new Date())).length : 0;
+  return { settings, updateSettings, connected: !!token && !!account, account, catalog, selected,
+    status, error, lastSynced, report, plannedRemovals, pending, conflicts, blockedWrites, multiUserEnabled,
+    connect, disconnect, preview, syncNow, resolveConflict };
+}
+
+
+// Preserve unrelated additions/edits queued while an async request was outstanding.
+export function applyPlannedList(current, snapshot = [], planned) {
```
krelltunez, line 225: This is the subtlest function in the PR, reconciling planned state against React state that may have moved while the request was in flight, and it's the one exported function with no direct test. The rest of the suite is thorough enough that this stands out.

It probably also belongs in `src/utils/` rather than inside a hook file. `CLAUDE.md` asks for pure utilities to live there with tests, and this one is pure.

src/components/SettingsModal.jsx
```diff
@@ -1963,6 +1964,9 @@ const SettingsModal = () => {
                       </>)}
                     </div>
 
+                    <TodoistSettings />
+                    <hr className={borderClass} />
```
krelltunez, line 1968: nit: I can see what `[&+hr]:hidden` on the section is doing, preserving this separator for the case where `TodoistSettings` returns `null`. Clever, but the net effect when it does render is that the Todoist section is the only one in the panel with no rule above it and none below.

Folding the component into the standard `collapsedSettings` pattern (see my comment on `TodoistSettings.jsx`) lets this just be an ordinary `<hr>` like its neighbours.

src/hooks/useTodoistSync.js
```diff
@@ -0,0 +1,237 @@
...
+      if (navigator.locks?.request) {
+        await navigator.locks.request('dayglance-todoist-sync', { ifAvailable: true }, async lock => {
+          if (!lock) throw new Error('busyOtherTab');
+          await execute();
+        });
+      } else await execute();
+      succeeded = true;
+    } catch (err) {
+      if (generation.current === epoch) {
+        setError(err.message === 'cancelled' ? 'networkError' : err.message);
```
krelltunez, line 166: nit: This makes `errors.cancelled` in `strings.js` unreachable. Every abort path ends up here and gets remapped to `networkError`, so the string is never displayed. Either surface `cancelled` or drop the key (and its Chinese counterpart).

src/todoist/core.js
```diff
@@ -0,0 +1,270 @@
...
+  return tests.length > 0 && tests[settings.match === 'any' ? 'some' : 'every'](Boolean);
+}
+export function mergeResponse(previous = {}, response) {
+  if (!response || typeof response.sync_token !== 'string') throw new Error('invalidResponse');
+  const full = response.full_sync === true || response.full_sync === 1;
+  const user = response.user ?? previous.user;
+  const next = { ...previous, cursor: response.sync_token, user: user ? { id: user.id, full_name: user.full_name, timezone: user.tz_info?.timezone || user.timezone } : null };
+  for (const resource of ['items', 'projects', 'labels']) {
+    if (!Array.isArray(response[resource])) throw new Error('invalidResponse');
+    next[resource] = Object.assign(Object.create(null), full ? {} : previous[resource]);
```
krelltunez, line 68: blocking: This cache grows without bound.

Each incremental response is merged into the previous cache and nothing is ever evicted. Completed and deleted items arrive in the delta with `checked` / `is_deleted` set and then stay forever, since a full snapshot only happens at connect time. After a year of daily use, `dg-todoist-state-v1:<id>` holds every task the user has ever completed.

This matters more than it looks. dayGLANCE is localStorage-first, and localStorage is capped at roughly 5 MiB per origin regardless of available disk, because it's synchronous and so never scales with the machine. My own install is already at 60% of that budget, which makes an unbounded new consumer a real hazard.

The fix needs to be retention-aware, since `mirrorRemoval` genuinely needs `is_deleted` records while a local copy still exists. Something like a `pruneCache(cache, retainIds)` here in `core.js`, called from the hook just before `persist()` with the currently linked task ids, dropping entries where `!active(item) && !retainIds.has(id)`. Completed items are never imported anyway, since `additions` filters on `active`, so unlinked ones are safe to drop.

src/components/TodoistSettings.jsx
```diff
@@ -0,0 +1,215 @@
...
+import { useSyncCtx } from '../context/SyncContext.jsx';
+import { active } from '../todoist/core.js';
+import '../todoist/strings.js';
+
+export default function TodoistSettings() {
+  const { t: todoistText, i18n } = useTranslation('todoist');
+  const { darkMode, borderClass, textPrimary, textSecondary } = useDayPlannerCtx();
+  const { todoist: sync } = useSyncCtx();
+  const [input, setInput] = useState('');
+  const [expanded, setExpanded] = useState(true);
```
krelltunez, line 14: Every neighbouring section in the settings modal (TRMNL, Obsidian, cloud sync) uses `toggleSettingsSection` with the persisted `collapsedSettings` state from `App.jsx`, and defaults to collapsed. This local `useState(true)` means the Todoist section is expanded by default and forgets its state between visits, which makes it the odd one out in a fairly long panel.

Switching to the existing pattern also resolves the separator issue I flagged in `SettingsModal.jsx`.

krelltunez (commented):
Thank you for this, and genuinely, thank you for the care that went into it! I ran the full suite (3,502 tests) and lint locally, both green, on top of CI.

Some things I want to call out specifically, because they're the parts that integrations like this usually get wrong:

- **Treating a missing record as "no information" rather than a deletion** (`core.js:108`, `core.js:212`). This is the classic sync data-loss bug. You didn't just avoid it, you wrote tests asserting it.
- **Failing closed when every filter is empty** (`core.js:59`), instead of silently importing an entire account.
- **Durable command UUIDs written before the request**, so a retry can never produce a second close.
- **Web Locks rather than an ad-hoc localStorage mutex**, and degrading to read-only when it's unavailable instead of pretending it worked.
- **`dueParts` separating Todoist's fixed-zone instants from floating wall-clock times**, and rejecting malformed dates rather than letting `Date` silently shift them.

Also: importing as `bg-red-500` to match Todoist's brand is exactly right. It mirrors how Obsidian tasks come in purple, and I confirmed a user recolor survives sync, since `reconcileTask` never touches `color`.

The 72 tests are aimed at failure modes rather than line coverage, and that's the main reason I'm comfortable taking this. Keeping `App.jsx` to seven lines and putting the logic in a hook plus a pure core module is exactly the direction the codebase is moving.

I want to merge it. I've left detailed comments inline on the relevant lines; here's the map.

**Blocking, both inline:**
1. The cached Todoist snapshot grows without bound (`src/todoist/core.js`)
2. `disconnect()` leaves that cache on disk with no way to purge it (`src/hooks/useTodoistSync.js`)

**Would like, inline:**
3. `applyPlannedList` needs a test, and probably belongs in `src/utils/`
4. The settings section should follow the existing collapse convention
5. A small separator inconsistency in `SettingsModal.jsx`

**Would like, one that isn't anchorable to a changed line:**
6. Add `dg-todoist-state-v1` to the `labels` map in `src/components/StorageBreakdownModal.jsx:17`, so the cache shows a friendly name in the storage breakdown rather than a raw key. Worth doing in the same pass as item 1, since that's where its size will become visible.

### Translations, I'll handle this one

`strings.js` inlines English and a complete Chinese translation and registers them at import time. The repo deliberately moved away from that. The comment at `src/locales.js:15-17` records that eagerly inlining locales pushed the main chunk past the PWA precache ceiling, so each language is now its own lazily loaded chunk. In the current build each locale is a separate chunk of roughly 95 to 101 kB, while these 16.8 kB sit in the main bundle for every user regardless of language. It also means de, es, fr, it, pt-BR and pt-PT fall back to English.

This is more about repo plumbing than about your feature, and I know the file layout better than the diff does, so leave it to me. I'll move the keys into the existing `translation.json` files under a `todoist` prefix after merge, and pick up the other six languages at the same time. Nothing for you to do here, I just want it recorded so it doesn't get lost.

### Nits

- A short `docs/todoist-integration.md` would match the Obsidian, MCP and Tasker precedent.
- The style is considerably denser than the rest of the codebase, `reconcileLists` and the `report.reason` chain especially. Not blocking, but I'll be maintaining it, so a little breathing room would help.

(One more nit is inline.)

### One question

Does mirror mode need to ship in v1? "Today's tasks" is the path most people want, and mirror mode carries most of the risk surface, since it's the only mode that moves tasks to the recycle bin. Deferring it would shrink what we're committing to support without touching the part users actually asked for. Not a demand, just want your read.

Lcub3d:
Thank you for the detailed review, and for taking the time to look at this integration so carefully.

### Why I built this

A little context on why I built the Todoist integration in the first place: to me, dayGLANCE is a very ambitious planning / todo app with a lot of potential, and I really like the overall direction of its design. In particular, the **Goal → Project → Task** structure is something I find much stronger than in Todoist. The local-first approach and integrations with local tools such as Obsidian are also things I really value, and they are quite different from the cloud-first model of a conventional commercial todo app.

That said, I am still a long-time Todoist user. Some parts of it are still extremely convenient for me, especially quickly capturing tasks with natural-language dates and quick priority/project/label entry, so at the moment I have not been able to leave it entirely.

Because of that, my original idea was not to turn Todoist and dayGLANCE into two fully mirrored systems. I mainly wanted a lightweight bridge between them: continue using Todoist for fast capture, then bring the tasks that actually need deeper planning into dayGLANCE based on project, label, priority, etc., and do the more deliberate planning there.

### On mirror mode

**Mirror mode is not a core requirement for me.** My actual use case is much closer to **selective import / filtered sync** than to maintaining strict equivalence between the two task sets. If mirror mode materially increases the data-loss risk, maintenance complexity, or long-term support surface, I am completely in favor of deferring it from v1 — or removing it from this PR entirely.

This PR was also a bit of an experiment on my side. I originally built the feature for my own workflow, and once it existed I thought it was worth submitting in case it might also be useful to other users or to dayGLANCE itself. Which parts belong in the main project should absolutely follow your view of dayGLANCE's direction and what is reasonable to maintain.

### Longer-term direction

For a bit more context on my longer-term personal workflow: I would like Todoist to act more as a fast capture tool, while using dayGLANCE for longer-term planning, important-task management, and eventually something closer to a PDCA loop — not only recording the Plan, but also the actual Do for the day, then comparing and analyzing Plan vs. Do as the Check step. The idea is that it could become both a planning tool and something closer to a journal of what I actually did.

I also realize that this goes well beyond dayGLANCE's current product direction. If I keep developing that idea, I intend to do it in a separate branch rather than asking the main project to accommodate those personal requirements. For now I am tentatively calling that direction `dayglance-jobo`.

So for this Todoist integration, please treat it as a feature that came from my personal workflow rather than something I need the main project to preserve in its current scope. In particular, mirror mode does not need to be kept for me, and I am happy for the mainline scope to be reduced in whatever way you think is safer and more appropriate.

I'll also work through the blocking items and the other review comments as I get time. Thank you again for the guidance, and for being willing to take this PR.

krelltunez:
Thank you for this, it's a genuinely helpful reply. I'm really excited to collaborate with you!

### Mirror mode

Let's take you up on that. Could you drop mirror mode from this PR? To be clear about why: it's the only path in the integration that moves a user's tasks into the recycle bin, so it carries most of the data-loss surface for what you've said isn't your actual use case. Removing it should also make your remaining work smaller, since a fair number of the review comments touch mirror code, and it simplifies several of the gnarlier conditionals along the way, including the empty-result reason chain.

Please treat this as deferred rather than rejected. The work isn't wasted, and once the import path has been in real use for a while it's a reasonable thing to revisit.

The blocking items still stand for the selective import path.

On the Chinese translation: I'm going to move the strings into the repo's normal `public/locales/` structure after merge, but your zh-CN work will be carried over as-is. A native speaker's translation of their own feature is better than anything I'd produce, and I'd rather relocate it than replace it.

### On fast capture

Since fast capture is the reason you're still in Todoist, it's worth knowing dayGLANCE has some of this already, and importantly it's deterministic and offline with no AI call:

- `!`, `!!`, `!!!` for priority
- `%30` for a 30-minute duration
- `~15:00` for a time
- `@9/15` or `@9-15` for a date (optional year: `@9/15/2026`)
- `$` for a deadline

Those are numeric or punctuation-based, so they work regardless of interface language. There's also a natural-language layer ("dentist tomorrow 3pm every 2 weeks") but its vocabulary is English only, so it won't help you in Chinese today.

That's a gap I'd like to close, and you're better placed than I am to do it. The date and time display is already localized; it's only the typed keywords that are English-bound, and the vocabulary lives in named constants at the top of `src/utils/quickAddParser.js`. A Chinese vocabulary layer would be a well-scoped follow-up if it interests you. No obligation, and not a condition of this PR.

Separately, while looking into this I found that `#tag` autocomplete never triggers for non-Latin tags. The tags themselves work fine, they're stored, displayed and filtered correctly, but the suggestion dropdown never appears while typing them, so you can't discover or complete an existing Chinese tag. That's my bug, not yours. Filed as #1614.

### On dayglance-jobo

I'd like to hear more, and not so you can justify a fork. If what you have in mind fits dayGLANCE, I may well want to build it, collaboratively if you're interested.

Part of it may be closer to existing behaviour than you'd expect. Completed tasks are already logged with timestamps into the Obsidian daily note and per-project logs, and focus mode keeps its own session log. So the Plan and the Do are both being recorded already. What's missing is the Check: comparing them and analysing the gap.

There's a reason that's an appealing target. The morning summary and evening reflection features go through an AI call, and AI is not available in dayGLANCE builds for China, so those features aren't something you can use at all. But a Plan vs Do comparison doesn't need a model. The data is already on disk and structured. Done deterministically, it would work everywhere, including for you and for every other user who can't reach the AI features. That makes it more appealing to me in mainline, not less.

So I'd rather hear what you're picturing before you go off and build it separately.

What does "jobo" stand for, incidentally?

And thank you for what you said about the Goal → Project → Task structure and the local-first direction. Those are the parts I've put the most thought into and the hardest to get feedback on, so it's good to hear they land. You're also the first person other than me that I know of who's putting the Obsidian integration to real use, which is its own kind of milestone. There may well be others I haven't heard from.

src/components/SettingsModal.jsx
Lcub3d: Fixed: removed `[&+hr]:hidden` and kept an ordinary `<hr className={borderClass} />` in `SettingsModal.jsx`. The Todoist section now uses the shared collapse state, so it no longer needs the adjacent-divider workaround. The settings regression suite and the latest upstream CI pass.

src/hooks/useTodoistSync.js
Lcub3d: Fixed: the current-generation error handler now preserves `cancelled` instead of remapping it to `networkError`, so both localized cancellation messages are reachable. Added a direct hook regression in `src/hooks/useTodoistSync.test.js` for the aborted-request path. Latest upstream CI passes.

src/todoist/core.js
Lcub3d: Implemented `pruneCache(cache, retainIds)` in `src/todoist/core.js` and call it before persistence from the hook. Unlinked completed/deleted tasks and deleted project/label records are dropped; inactive task records referenced by local calendar/inbox/recycle-bin copies or pending commands are retained. Core and hook tests cover pruning and retention. Mirror removal/restoration is now absent entirely.

Also addressed the storage-breakdown item from the review summary: account-qualified cache keys display a friendly label. Commit `a49fe7d` loads that label from all eight normal lazy locale bundles and adds eight standalone rendering regressions; the translation-key checker remains unchanged. Upstream CI now passes all 3,693 tests. This limits accumulated inactive history, not the size of an arbitrarily large active account, which still shares the localStorage budget.

src/components/TodoistSettings.jsx
Lcub3d: Switched to `collapsedSettings.todoist` and `toggleSettingsSection('todoist')` from the app context, with the Todoist section collapsed by default. Removed the local expanded-state toggle. The mobile `variant="page"` remains accessible independently of the desktop section's state. `TodoistSettings.test.jsx` covers the collapsed/expanded/page behavior; the latest upstream CI passes.

src/hooks/useTodoistSync.js
Lcub3d: Adding the implementation/test details for this already-resolved thread: `disconnect()` now cancels in-flight work, clears session credentials, disables automatic sync/writeback, and waits for the account's Web Lock before clearing an empty-queue cache. If receipts remain, their original UUIDs and the pending count are preserved; reconnecting the same account can resolve them before a later disconnect clears the cache. Imported dayGLANCE tasks are not removed.

Direct hook regressions cover idle-cache removal, pending-receipt retention, corrupt-queue fail-closed behavior, and an in-flight read not recreating the cache after disconnect. These pass in the latest upstream CI.

src/hooks/useTodoistSync.js
Lcub3d: Moved `applyPlannedList` into `src/utils/todoistReconciliation.js`, imported by the hook. Its adjacent test file now has nine direct cases: unchanged-state fast path, concurrent additions/edits/deletions, no resurrection, removal of untouched rows, duplicate prevention, numeric/string ID normalization, and immutable inputs. The tests pass locally and in the latest upstream CI. Also expanded the dense `reconcileLists` and empty-result reason logic and added `docs/todoist-integration.md`, addressing the accompanying readability/documentation notes.

Lcub3d:
The review follow-up is ready for another look. I've replied with implementation/test details on all six inline threads and updated the PR description to match the current three-mode scope. Mirror mode and automatic local removal/restoration are deferred, as requested.

The last failing CI assertion was the storage-breakdown label using a key outside the normal lazy bundles. `a49fe7d` fixes that label in all eight bundles and adds standalone rendering tests, without changing the translation checker or moving the rest of the Todoist namespace. Your planned post-merge translation migration is still left to you.

Latest validation:

- [CI](https://github.com/krelltunez/dayGLANCE/actions/runs/34683978461): all 3,693 tests, lint, dependency audit step and production build passed.
- [iOS build](https://github.com/krelltunez/dayGLANCE/actions/runs/34683978440): passed.
- Local focused suite: 193/193; web, Android WebView and Electron renderer builds passed. The PR description records the 37 failures in the unfiltered local Windows run; I am not claiming a fully green local run or a live Todoist account end-to-end test.

Vercel still reports "Authorization required to deploy"; I have not changed deployment permissions. Thanks again for the detailed review and for helping narrow this to the selective-import path.

Lcub3d:
Thank you for the thoughtful response and the invitation to collaborate. I’m glad you’re interested in discussing this direction together.

### Mirror mode and translations

Agreed—mirror mode is now removed from this PR, and I’m treating it as deferred rather than rejected. I’ve posted the review fixes and validation results separately above, so I won’t repeat those details here.

Thank you also for preserving the Chinese translations. I’m happy to follow the translation migration plan you suggested.

### How I use Todoist

Your explanation of dayGLANCE’s offline quick-entry syntax was helpful, and thank you for identifying the non-Latin tag autocomplete issue.

I should clarify something from my earlier reply: I probably put too much emphasis on quick capture. Todoist’s task organization, scheduling, mobile experience, and ecosystem also fit my workflow well, so I’m not looking to replace it immediately.

**I’d like the two tools to complement each other.** Todoist can remain my primary task-management and planning tool, while I bring selected tasks into dayGLANCE for time-blocking, recording what actually happened, and reviewing my day. That’s a personal workflow preference, not a suggestion to weaken dayGLANCE’s own planning features.

A Chinese vocabulary layer sounds like a useful, well-scoped follow-up. I agree it should stay separate from this Todoist PR.

### What I mean by Jobo

“Jobo” isn’t an acronym—just a working name combining an association with *job* and a little of the sound of the Chinese word “账簿” (*zhàngbù*, meaning an account book or ledger). The idea is a ledger of actions and time.

I was probably too quick to describe it as outside dayGLANCE’s direction. A separate branch was meant to keep experiments from disrupting existing functionality, not to commit to a permanently separate fork. I’d be happy to discuss the needs first and explore what fits naturally into the main project.

Here’s a concrete example:

- In the morning, I plan to write a report from **09:00 to 10:00**.
- In reality, an unplanned work conversation takes **09:00–09:30**, and I work on the report from **09:30–10:00**.
- The report is still unfinished, so I move the remaining work to **16:00** and complete it at **16:45**.

The completion timestamp is useful, but it doesn’t by itself tell me how long I worked on the report earlier, how much time the interruption took, or what the original plan looked like before I rescheduled it.

What I’d like to preserve is the distinction between **the plan, actual time spent, and task completion**. One task might have several actual work intervals across multiple days without being complete yet.

### A possible interaction

I’m leaning toward semi-automatic recording:

- When a block happened as planned, confirm it as an actual record.
- When reality differed, adjust the actual interval or add an unplanned activity.
- Keep the original plan available for comparison, even after rescheduling the remaining work.

On the timeline, actual records could be prominent in the past portion, with the original plan as a lighter reference. The future portion would remain a planning space.

Your comments about Obsidian completion logs and focus-session logs are a good starting point. I’d want to understand and reuse those before proposing a separate recording system. The question is whether they already capture the intervals and original-plan history needed for that comparison, and what small gaps remain.

**The first version doesn’t need AI.** Recording the facts clearly and reviewing plan versus actual offline would already be valuable. I’d also want this to remain connected to the existing areas, goals, projects, and tasks—not become a detached time tracker.

This is a separate discussion from the Todoist PR, with no additional scope proposed for that PR. Does the example above fit how you envision the existing timeline and logs evolving, and where would you see the smallest useful first step?

krelltunez:
This has been merged and will ship in the next release (which I'm planning for this weekend)! Thank you for this. I want to be specific about what made it a pleasure to review, because it's rare.

You took a review that asked you to delete an entire feature you'd built, and came back within hours having done it, plus every other item, several of them done better than I asked for. A few things stood out:

- The retention set for `pruneCache` is more careful than what I suggested. I said linked task IDs; you worked out that recycle-bin copies and pending command receipts need their source records retained too.
- You found and closed a race I hadn't spotted, where an in-flight sync could re-persist the cache after disconnect cleared it. Taking the writer lock in `disconnect()` is the right fix.
- Passing `recycleBin` into the hook without `setRecycleBin` turns "this shouldn't remove tasks" into "this cannot remove tasks". That's the strongest form of the guarantee and I didn't think to ask for it.
- The migration for anyone who had the experimental mirror mode enabled, with automation and writeback forced off rather than silently re-pointed, is exactly the right instinct.

I also appreciate that you kept the Windows disclosure in the PR description after CI went green, and that you never claimed a fully green local run or an end-to-end test you hadn't done. That kind of precision is worth more than it probably feels like when you're writing it.

There's follow-up work on my side before release, mostly plumbing: moving the Todoist namespace into the normal locale bundles (carrying your zh-CN across as-is), a `CLAUDE.md` note for `src/todoist/`, and an issue for the active-account storage limitation you flagged. None of it is yours to do.

## dayGLANCE Jobo

I'm adopting the name. "dayGLANCE Jobo" it is. 账簿 crossed with "job", a ledger of actions and time, is a better name than anything I'd have come up with, and the fact that it means something in your language and not only in mine feels right for where this project is going.

Your example was the clarifying thing, so let me answer your actual question: what already exists, and what doesn't.

**Actual intervals already exist.** The focus log doesn't only store durations. It records a span per session in `src/App.jsx:4250`:

```js
const span = { start: startMin, end: startMin + sessionMinutes };
```

with a comment that reads like it was written for this conversation: "WHEN the session ran, not just how long, the Day Dial places it against the block it happened in." `computeFocusSpans` in `src/utils/dayDial.js:497` already merges and clips those, and the Day Dial already draws them against the plan. Up to 60 spans per day are kept.

So half of what you want is built. Real intervals, already rendered against intention.

**Three gaps in it.** Spans are session-level rather than per-task, so tasks carry an aggregate `focusMinutes` but not their own intervals. Only time inside Focus Mode is captured, so your unplanned 09:00 to 09:30 conversation wouldn't appear at all. And the spans are a flat per-day list.

**Original-plan history does not exist anywhere.** I went looking for any preservation of a prior schedule and there is none. Rescheduling overwrites `date` and `startTime` outright. So in your example, the moment the report moves to 16:00, the fact that it was planned for 09:00 to 10:00 is gone permanently.

**That makes the smallest useful first step clear**, and it's smaller than either of us was imagining: preserve the original plan. Either a first-planned `{date, startTime, duration}` on the task, or a short append-only list of schedule revisions, written at the existing reschedule sites.

The reason to do that one first is that it's the only piece that cannot be reconstructed later. Every day we don't record it, that history is destroyed. The comparison view, the timeline treatment, per-task intervals, semi-automatic confirmation: all of those can be built afterwards on data we've been accumulating. This one can't. It's also additive, needs no UI to start being useful, and once it's there the Day Dial already has both halves of the comparison to draw.

## The bigger question I'd like your view on

Thinking about your ledger idea gave me a framing I like: **dayGLANCE as the record of intention, and Obsidian as the record of what actually happened.**

Here's the catch, and I think it's genuinely interesting rather than a problem. That split is currently half-built and half-inverted. Completion logs already flow out to the Obsidian daily note and per-project logs, which fits. But the focus spans, the truest record of what actually happened, live in `localStorage` and are drawn on the Day Dial, which doesn't.

So: should actual intervals flow to Obsidian too, making the vault the durable ledger, with dayGLANCE the live instrument for seeing plan against actual? Or is the Day Dial the right home for them and the vault should stay a completion record?

I don't have a settled answer and I'd rather work it out with you than decide it alonYou're actually running this loop day to day, comparing what you planned against what happened, which I haven't been doing in any disciplined way. So your instincts on what matters here are worth more than mine.

One practical note: this discussion will outlive a merged PR thread. I've opened [#1623](https://github.com/krelltunez/dayGLANCE/issues/1623) with what I found and the open question, so it has somewhere to live beyond this merged PR.

Thanks again. This was a genuinely great contribution and I'm glad you submitted it.

Issue #1623, opened 2026-09-12
[Feature] dayGLANCE Jobo: a ledger of plan versus actual (Planning)


krelltunez:
Design discussion, seeded from a conversation with @Lcub3d on #1604. Not a committed roadmap item yet.

## The idea

dayGLANCE records what you *intended* to do. It does not yet record what you *actually* did, in a way you can compare the two.

"Jobo" is @Lcub3d's working name, combining "job" with the sound of 账簿 (*zhàngbù*), an account book or ledger. A ledger of actions and time. Adopting it as **dayGLANCE Jobo**.

## The motivating example

From @Lcub3d, lightly reformatted:

- In the morning, plan to write a report from **09:00 to 10:00**.
- In reality, an unplanned work conversation takes **09:00 to 09:30**, and the report gets worked on from **09:30 to 10:00**.
- The report is still unfinished, so the remaining work moves to **16:00** and completes at **16:45**.

A completion timestamp alone does not tell you how long you worked on the report earlier, how much the interruption cost, or what the original plan looked like before you rescheduled.

The distinction worth preserving is between **the plan**, **actual time spent**, and **task completion**. One task may have several actual work intervals across several days while still not being complete.

## What already exists

**Actual intervals are already recorded.** The focus log stores a span per session, not just a duration. `src/App.jsx:4250`:

```js
const startMin = focusSessionStart.getHours() * 60 + focusSessionStart.getMinutes();
const span = { start: startMin, end: startMin + sessionMinutes };
```

The comment there already states the intent: "WHEN the session ran, not just how long, the Day Dial places it against the block it happened in."

`computeFocusSpans` (`src/utils/dayDial.js:497`) merges and clips them, `focusSpanMinutes` (`:542`) totals them, and the Day Dial renders them against the planned blocks. Up to 60 spans per day are retained (`FOCUS_SPANS_PER_DAY`, `src/App.jsx:4167`).

**Completions already flow to Obsidian.** `src/hooks/useCompletionLog.js` writes completed tasks with timestamps into the daily note and per-project logs.

**Per-task focus totals exist.** Tasks accumulate `focusMinutes` (`src/App.jsx:4235`), though as an aggregate rather than as intervals.

## What is missing

[...]

PR #1673, opened 2026-09-16
feat: opt-in Jobo plan/do journal and PDCA prototype

Lcub3d:
Refs #1623. Draft for design and implementation review, not a request to close the design issue or ship the entire prototype immediately.

@krelltunez, this answers your two questions: yes, preserve a first-observed plan baseline; and keep the daily record/correction/comparison loop in dayGLANCE, with Obsidian as an optional later archive rather than a required backend. Todoist can remain the primary task manager.

Full design, exact decision rules, two Mermaid diagrams, limitations and source review order: https://github.com/Lcub3d/dayGLANCE/blob/dayGLANCE-jobo/docs/jobo-design.md

## What is implemented

This is a normal source port against main a39d73c9, not the earlier Windows compiled-renderer packaging. One clean commit, 26 source/docs/locale/test files; no binaries, workflows or new dependencies.

Experimental desktop Settings switch, OFF by default. Native DAY/WEEK and checkbox semantics remain when off; MULTI/MONTH/SCHED are unchanged. No new mobile Jobo layout.

When enabled: native-themed Plan | Do | Notes, one observed baseline with mutable current plans, separate execution intervals, shared native notes, compact badges and connections. Checking an unchecked plan appends a default-completed Do ending now and starting one full planned duration earlier. This is an editable self-report, not measurement or native/Todoist task completion. Partial/Started unchecks; another check appends a new attempt. Only the most recently appended attempt's progress appears on Plan, not an aggregate.

Timing and progress are independent. Delayed compares interval position to the baseline; Overrun compares summed duration; Interrupted means two or more records and can coexist with both. An entirely elapsed CURRENT displayed plan without Do is Not started in the ledger, not evidence of unrecorded real-world behaviour. Future plans are not labelled. Unplanned Do does not get an invented budget. Overlap is currently summed, not unioned.

WEEK shows Do on dates before today and Plan today onwards, with compact status/progress colours. Carry Forward copies user-selected unfinished work to new native inbox tasks with provenance/deduplication, without deleting history or copying remote identities.

## Release gaps and review request

Baseline capture is not universal across all scheduling paths. The ledger backend is still localStorage; it has an injectable fail-closed store and explicit JSON backup, but IndexedDB, atomic cross-tab writes and integration with every native restore/reset/cloud path are NOT finished. Import is not transactional across native and ledger collections. No Do sync to Todoist/Obsidian is added; ordinary native note/task integrations keep their existing behaviour.

Recurring/provider identity, DST/timezone travel, mobile and full accessibility audits remain. Focus-log import, Day Dial/MONTH comparison, statistics and AI are deferred. New UI copy is English/zh-CN; other bundles use explicit English fallback. A smaller baseline/core-first split is welcome.

## Validation

https://github.com/Lcub3d/dayGLANCE/actions/runs/35070089496 passed lint, all 4,087 tests in 291 files and production web build, then pushed source commit 4fdf3b9. This is fork validation, not later upstream PR CI.

Local: 133 new tests, web/Android WebView/Electron renderer builds and 15 offline source-built Chromium smoke checks passed. The unfiltered local Node22.16 suite had the same two German Intl duration-format failures on clean main (3,952 pass/2 fail) and this branch (4,085 pass/2 fail); no exclusions. No Windows launch, real provider round trips or production migration audit is claimed.

Prepared with ChatGPT assistance. Maintainer edits are welcome.


Lcub3d:
## Context and answers to #1623

Refs #1623; follows our discussion in #1604. This is a **draft for design and implementation review**, not a request to close #1623 or replace everyone's day planner immediately.

@krelltunez, thank you for following up and for the storage work. My answers to your two questions are:

1. **Yes, preserving the original plan is the right foundation.** This prototype retains one first-observed timed baseline while allowing the current plan to move. It does not pretend to reconstruct schedules from before recording began. Capture is not yet wired into every upstream rescheduling path; extracting that smaller, universal baseline change first remains a reasonable merge strategy.
2. **I would keep actual intervals in dayGLANCE, with Obsidian as an optional archive rather than a prerequisite.** Recording, correcting and comparing the day should work without a vault connection. A readable Obsidian archive would be valuable later, but I would avoid two independently editable sources of truth in the first iteration. The current code does not export Do intervals to Obsidian.

I explored the visible interaction as well as the baseline so there is something concrete to evaluate. This is deliberately a larger prototype than your proposed first step, not an assumption that the whole design should land together. The behaviour below comes from developing and trying the workflow, not from a claim that I already have a mature, long-running quantified-self practice.

## Product idea: task continuity × time

A task is ongoing work that can remain connected to an area, goal or project across days. A plan says when I intended to work; a Do record says when I report actually working. Completing a task is a third fact, not something to infer from either interval.

Todoist can remain my primary task manager and planner, while dayGLANCE imports selected tasks and helps me record and review the day. Native dayGLANCE planning remains available. The aim is a time journal connected to tasks, not another Todoist replacement or a detached stopwatch.

The paper-planner inspiration is the side-by-side relationship between intention, execution and notes. The implementation reuses the existing dayGLANCE shell, theme, task content/actions, native note editor, inbox and week layout. No paper-planner assets or fonts are included.

**This PR is a normal React/ESM source port against `main` at `a39d73c9`.** Earlier local Windows previews were used to iterate on the interaction; their installers, compiled-renderer patching, ASAR files and launchers are not in this diff. The source branch has one commit and 26 changed files, with no dependency or native build changes.

## Try it without changing the default experience

Enable the experimental **Jobo** option in desktop Settings under view preferences, then open DAY. It is **off by default** and only mounts after data has loaded in a single-user, non-tray context.

While disabled, the original DAY/WEEK and native checkbox semantics remain. MULTI, MONTH and SCHED are not replaced. Mobile does not gain a Jobo layout in this prototype. Disabling Jobo hides it without deleting its ledger.

## DAY: Plan | time ruler | Do | Notes

The headings follow the selected language, rather than displaying two languages together. DAY keeps the native background and task colours. Cards use two compact rows: title/actions, then time range, duration and applicable badges.

- **Plan** remains freely reschedulable. Its captured baseline remains separately available for comparison.
- **Do** is a separate record with its own ID, interval and progress. Dragging the same plan into Do multiple times creates independent attempts; adjacent attempts are not automatically merged.
- Linked Do cards show title, time and actions, without repeating the badge row. The Plan card carries the comparison and **one progress control for the most recently appended attempt**. Unplanned Do has no Plan counterpart, so its own card carries Unplanned, progress and custom tags.
- Progress is a content-width badge button, without a dropdown arrow. Time, progress and grey `#` tags share the same compact treatment. `+` opens a small input; blur/Enter commits, Escape cancels.
- The original Plan note button still opens the native notes/subtasks editor. Only Do's note button adds a right arrow and opens/focuses a Notes tile. Both edit the same canonical native task note; independent legacy note text is archived before shared-note migration.
- Notes stack vertically, independently of time. Nonempty notes and explicitly opened blank notes appear, followed by the existing daily note. They are directly editable and resizable; resizing changes display height, never duration. Deleting a note does not delete the task or Do.
- Hover/drag shows Plan–Do–Notes connections. Multiple Do records for a plan have a persistent square bracket at the right. The current-time line stays above cards and connectors without intercepting input.

## Low-friction recording and the checkbox contract

Within Jobo only, checking an unchecked Plan means **record an execution attempt**, not complete the native task or send Todoist completion writeback.

The proposed interval ends at the current minute and extends backwards by the current full planned duration. For example, checking a 30-minute plan at 14:20 records 13:50–14:20. This is an editable retrospective default, **not measured activity and not proof of uninterrupted work**. Moving, resizing or editing the interval corrects it.

Each new Do defaults to Completed to keep the common operation to one click. The choices are Started, Partially completed, Mostly completed and Completed. They describe the user's assessment of that attempt, not a numerical task completion percentage.

A Plan displays just the latest **appended** attempt's progress, not an aggregate and not whichever interval has the latest edited timestamp. Older attempts retain their individual progress. Changing the visible attempt to Started/Partially completed unchecks the Plan; checking again appends another default-completed interval ending now, with the full planned duration. It does not overwrite the earlier record or estimate remaining effort. Mostly/Completed keeps it checked. Clicking an already checked Plan does not accidentally append another record; dragging remains the explicit repeat-record action.

This checkbox meaning is intentionally isolated behind the feature switch. It is one of the design choices I would especially like feedback on before general availability.

## Check part 1: timing and progress are independent

| Timing badge | Prototype rule |
| --- | --- |
| Within plan | There is linked Do, with neither delay nor overrun |
| Delayed | Earliest actual start is after baseline start, or latest actual end is after baseline end |
| Overrun | Sum of linked Do durations exceeds baseline duration; gaps are not counted |
| Interrupted | Two or more linked Do records, including adjacent records; independent of the other badges |
| Not started | No linked Do, and the **current displayed plan has entirely elapsed** |
| Unplanned | No corresponding timed baseline; no comparison against an invented budget |

Delayed, Overrun and Interrupted may coexist. Interrupted may also coexist with Within plan. It is a record-segmentation heuristic, not automatic detection of a real interruption or its cause.

For Not started, a future or currently straddling plan is not labelled, and an old baseline must not incorrectly label a rescheduled future block. By deliberate workflow choice, no record after the displayed plan ends means not started **in this ledger**; there is no separate forgot-to-record state. Future analytics must not confuse that convention with evidence about unrecorded real-world behaviour.

Overlapping intervals are currently summed as entered, not unioned into wall-clock coverage. That tradeoff is explicit; a future statistics view needs its own overlap policy. Timing never changes native task completion, and progress is never aggregated across attempts.

## WEEK and Act

WEEK retains the native seven-column grid. Entire dates before today display Do; today and later dates display Plan. It does not fill missing actuals with plans. Compact colour edges represent timing, while a progress dot/text emphasis represents the individual attempt. Completed backgrounds recede toward the grid without hiding the interval or its deviations. Ctrl+wheel adjusts vertical scale; ordinary wheel scrolls.

**Carry Forward**, beside New task in the inbox, offers elapsed plans with no Do and individual Started/Partially-completed attempts. The user selects candidates; confirmation copies new native inbox tasks with provenance and deduplication, leaving the original plan and Do untouched. External identities, credentials and completed state are not copied. Historical partial attempts may remain candidates after a later successful attempt: the user decides rather than the app inventing total completion.

```mermaid
flowchart LR
    P[Plan: organise work and retain a baseline] --> D[Do: checkbox, drag or unplanned entry]
    D --> C[Check 1: independent timing and attempt progress]
    C --> A[Act: select work to carry into the inbox]
    A --> P
    C --> N[Notes and intervals retained for review]
    C -. later .-> R[Statistics and deeper human or AI review]
```

The full rules, a second timing decision diagram, data ownership and review order are in [`docs/jobo-design.md`](https://github.com/Lcub3d/dayGLANCE/blob/dayGLANCE-jobo/docs/jobo-design.md).

## Storage and release boundaries: intentionally not hidden

This prototype treats the ledger as original user data, not a reconstructible cache. Records are not automatically pruned when a source task is deleted or rescheduled.

The storage wrapper is injectable and lazy. Corrupt reads fail closed; unsuccessful quota writes/undo are not acknowledged as successful state changes; storage events and a stale-snapshot check catch detected conflicts. **That is not an atomic cross-tab transaction.**

**The ledger backend is still localStorage, not IndexedDB.** Your new helper gives the long-lived ledger a sensible next storage path, but it would be inaccurate to claim that migration is implemented here. Quota, multi-tab concurrency and native backup/restore/reset lifecycle need a coherent solution before release. Being default-off is not a substitute for that work.

An explicit Jobo JSON backup covers ledger records/baselines/progress/tags/presentation metadata plus native tasks, inbox and daily notes. Import validates and confirms before replacing those local collections. It is **not transactional across the native and ledger stores, and is not integrated into every native backup/cloud/file/reset path**. Restored native tasks and native note edits can still participate in integrations already enabled by the user. Do records themselves have no new remote sync or deletion path.

Other limitations are documented: baseline capture currently observes enabled DAY/WEEK and Jobo operations rather than every scheduling path; recurring-series/provider identity semantics, DST/timezone travel, mobile and comprehensive accessibility need further work. Existing focus spans are not imported or automatically attributed to tasks. Day Dial and MONTH comparison, automatic tracking, statistics/AI and Obsidian interval export are deferred.

English and Simplified Chinese new strings are supplied through the normal locale bundles. The other six bundles contain explicit English fallback copy for the new namespace, not six claimed reviewed translations.

## Validation

The [source publication validation run](https://github.com/Lcub3d/dayGLANCE/actions/runs/35070089496) passed **dependency installation, lint, all 4,087 tests in 291 test files, and production web build**, then pushed exactly `4fdf3b9` to this branch. This is fork validation, not a claim about later upstream PR checks.

Additional local checks:

- All **133** new pure logic, feedback/linkage, preference and store tests passed.
- Web, Android WebView and Electron renderer production builds passed.
- **15** source-built, offline Chromium smoke checks passed, covering default-off/enable/disable, native views, three columns, timing/progress, append-at-now, checkbox isolation, custom tags and unplanned Do. These used synthetic tasks and in-memory browser storage, not a real account or a native Windows session.
- `git diff --check` and source lint passed. No tests or translation checks were disabled.

For completeness, unfiltered local Node 22.16 runs on the same environment gave:

| Checkout | Passed | Failed |
| --- | ---: | ---: |
| Clean `main` at `a39d73c9` | 3,952 | 2 |
| This source port | 4,085 | 2 |

The two failure identities matched: a German duration expectation in `DayDial.a11y.test.jsx` and one in `localeFormatting.test.js` (`1h` versus this host's `1 Std.`). Remote full-suite validation passed. I am retaining the local disclosure rather than describing every environment as green.

No source-port Windows installer launch, real Todoist/Obsidian/cloud round trip, full mobile/DPI/accessibility sweep or production migration audit is claimed.

## Suggested review order / what I am asking

Start with `src/utils/jobo.js` and tests, then `src/jobo/store.js` / `preference.js`, `JoboRoutes.jsx`, and finally the views/scoped CSS and small native integration seams.

I would especially value agreement on baseline ownership, the retrospective checkbox default, per-attempt rather than aggregate progress, and keeping the ledger local with optional archival. A baseline/core-first split is welcome; this draft is here to make the complete interaction and its remaining risks inspectable, not to turn the exploratory design into a settled roadmap.

Prepared with ChatGPT assistance. Maintainer edits are welcome.

krelltunez:
I have been running this with the demo data, and it is genuinely impressive. A few things stood out immediately:

- Completing a task appending a Do entry automatically, already tagged on the timing and progress dimensions, so the record is a by-product of working rather than a second thing to maintain.
- Changing Completed to Partially Completed to re-open a task for a second session. That is the right shape for how work actually goes.
- Everything landing in an exportable JSON ledger. That is a good decision beyond the immediate feature, since it is the kind of record that is worth analysing later.
- Shipping demo data with it. That made evaluating this take minutes instead of an evening.

Before anything else, a correction on our side. On #1623 I said the schedule history popup showing only the original plan and the current one was by design. That was true of what existed, but your point stood, and we have since built the intervening moves: #1713 added a count of how often a task slipped after it came due, and #1724 records where each of those slips landed, so the popup now shows the route rather than two endpoints. Your framing changed our mind, and it seemed better to say so than to quietly ship it.

That is part of a broader thing worth being direct about. Since this PR opened, main has moved: the first-observed plan baseline you asked for now exists (`originalPlan`, write-once, carried through both sync tiers), along with the slip count and the trail. So some of what this branch carries has landed independently, and the branch is now against a fairly old main.

## How we would like to proceed

We do not think this should be merged as one piece, and we want to be clear that is not a rejection. A 26-file change against a main that has moved, adding a second data store, is hard to review and harder to keep alive. What we would rather do is land it in main incrementally, behind a flag that is off by default, so that you and we are both building on the same base instead of maintaining a fork that drifts.

The app already has the machinery for this:

- Optional features use a persisted `xEnabled` flag with a Settings switch (`habitsEnabled`, `routinesEnabled`, `goalsProjectsEnabled`).
- `src/constants/views.js` is a data-driven registry. Adding a view means appending it to `DESKTOP_VIEW_MODES`, giving it a shortcut key and a label key. The keyboard handler, the view cycler and the shortcut sheet all read those maps, so they need no changes. `enabledViews` and `resolveStoredView` already handle a view being off on a device, including a persisted value arriving from a device where it is on.

## The shape of the first slice

So that anything you start now lands cleanly, this is what the scaffold will look like. A PR will follow shortly, but the names are settled:

- A `joboEnabled` flag, off by default, with a switch under an Experimental section in Settings. Same shape as `habitsEnabled` and `routinesEnabled`.
- The view is registered as `'jobo'` in `src/constants/views.js`, on key `6`, labelled JOBO. With the flag off it is treated as a hidden view, so the existing machinery keeps it out of the cycler, the number keys and the shortcut sheet, and a persisted `'jobo'` arriving on a device where it is off falls back safely.
- `src/components/JoboView.jsx` renders when the view is active. In the first slice it is a placeholder.
- The ledger will live in `src/jobo/`, following the split the Todoist integration uses: `core.js` for pure rules with tests, a separate module for persistence and sync, and a hook wiring them to React state. Nothing in the first slice touches data.

If you want to start on something in the meantime, the most valuable piece is the pure part of the ledger: the record shape, the status transitions, and the timing and progress classification, as functions with no storage and no React behind them. That can land into `src/jobo/core.js` regardless of what the persistence ends up looking like, and it is the part of the prototype you have already tested most thoroughly.

## The order that matters

We would like the ledger to land before any of the interface.

You have been straight about where it stands: localStorage only, no IndexedDB, no atomic cross-tab writes, not integrated with every restore, reset and cloud path, and import not transactional across the native and ledger collections. We take that seriously because we have just been through a much smaller version of it. `originalPlan` is a single field, and it shipped in a state where it was written to localStorage but never reached React state, so it could not survive one sync cycle. It took a working install reporting 0 of 626 tasks with a baseline to notice.

There is now a checklist in CLAUDE.md under "Adding a field to a task" that came out of that, and it applies to the ledger too: does gaining the data count as a user edit, do both transports carry it, does the apply carry it, and does it reach React state rather than only storage. A new store that answers those wrongly loses records rather than losing a field, so we would rather settle it with no interface attached.

## Design feedback

Four things that affect what gets built rather than how it looks.

**JOBO should be its own view, not a replacement for DAY.** Our instinct is a sixth view alongside MULTI, DAY, WEEK, MONTH and SCHED, reachable on 6, with the existing views left as they are when the flag is off.

**Your WEEK behaviour is the part we are most excited about, and we think you already had it right.** This PR describes WEEK showing Do on dates before today and Plan from today onward, and on #1688 you noted the prototype does not yet follow that rule. Past days becoming a record of what was actually done, rather than a row of checked boxes, is a better argument for this whole idea than the DAY view is. We would like to see that working.

**The Notes column is lovely on a wide screen and probably should not be required.** The app already has notes panels, so JOBO working on a narrower viewport without it seems worth keeping possible.

**Carry Forward overlaps with something we have open.** Task cards already have a single-click send-to-inbox action, so the button is mainly a bulk version of it, which is useful. It also overlaps #1714, on generating successor tasks, which is squarely the Act in PDCA. Worth designing those together rather than separately.

One smaller thing: JOBO renders tags differently from the rest of the app and adds a `+` button for adding them, which does not exist elsewhere. Either is defensible, but having the mechanic in one view only is confusing. We would either drop it and keep tag entry in the task title, or adopt it app-wide.

We have a further list of polish notes, on menu and popup consistency with the rest of the app, the daily note's placement and prominence, collapsible note cards, and undo without a matching redo. We are holding those rather than sending them now, because they are the wrong thing to be discussing about a prototype you have labelled a prototype. They will make more sense slice by slice.

## What we would ask

If this way of working suits you, we will open a tracking issue with the slices, land the flag and the empty view ourselves, and then take the ledger. You would be very welcome to take any of the slices instead, against main rather than a fork. We will keep this PR open as the design reference, since the design document and the prototype are the most useful things anyone has written about this idea.

#1675 we would hold until the desktop shape settles, so you are not maintaining two prototypes at once.

Lcub3d:
This approach works for me.

I agree with splitting the prototype into smaller slices against `main`, and I'll take the pure ledger/core slice in `src/jobo/core.js`: the record shape, status transitions, timing/progress classification, and tests, without persistence or React wiring.

The scaffold and the proposed separation between core, persistence/sync, and the view also make sense to me. I'll keep the core independent so the later slices have a stable contract to build on.

Thanks also for the detailed feedback on the prototype. I'll treat #1673 as the design reference rather than trying to keep the whole branch current with `main`.

krelltunez:
The first slice (#1725) has been merged into `main`. This includes the joboEnabled flag, JOBO seated as a sixth desktop view on key `6`, and a placeholder where the view will go. The names match what is described above, so anything built against them will land cleanly. Nothing in it touches data; the ledger is the next slice, and src/jobo/core.js is yours to begin.

PR #1699, opened 2026-09-17
Fix weekly recurrence limits after fast-forwarding

### LINKED ISSUE #1698: Weekly recurrence can exceed its occurrence limit in later date ranges
I run Live Tennis API.

A weekly or biweekly task with an occurrence limit can appear again after the series has ended, depending on the date range being viewed.

Reproduced on main at `c48371bf`, using the shared recurrence module on Linux with Node 22.23.2:

```js
import { getOccurrencesInRange } from './packages/agenda-core/src/recurrence.js';

const task = {
  recurrence: {
    type: 'weekly',
    startDate: '2026-08-03',
    daysOfWeek: [1],
    maxOccurrences: 3,
  },
};

getOccurrencesInRange(task, '2026-08-03', '2026-09-30');
// ['2026-08-03', '2026-08-10', '2026-08-17']

getOccurrencesInRange(task, '2026-08-24', '2026-09-30');
// Actual: ['2026-08-24']
// Expected: []
```

The later window should agree with filtering the full series. The weekly fast-forward path currently omits the first skipped cycle from its consumed-occurrence count. It also affects multi-day biweekly series and can cause `getNextOccurrence` to report another occurrence for an exhausted series.

I have a focused fix and regression cases for full/partial initial weeks, skipped/deleted dates and next-occurrence lookup. Ten cases fail against the current implementation in the project's Vitest harness.

bensynapse:
I run Live Tennis API.

Fixes #1698. A weekly task starting August 3 with a three-occurrence limit ends on August 17, but querying from August 24 currently returns a fourth occurrence. The fast-forward path now counts eligible weekdays in the initial partial week and all subsequent skipped cycles, so later views respect the same limit as a full-series query. Deleted and skipped occurrences continue to consume the limit.

Added 12 cases covering weekly/biweekly series, partial or empty initial weeks, default weekdays, exceptions, DST-season dates and next-occurrence lookup. Ten fail on the original implementation.

Validation on Node 22.23.2:

- `TZ=UTC npm test`: all 4,313 tests pass across 294 files.
- `TZ=America/New_York npm test -- src/utils/recurrenceEngine.test.js`: all 52 tests pass.
- `npm run lint`, `npm run audit` and `npm run build` pass. The audit reports only the project's already-triaged advisories.
- The unchanged Dockerfile builds successfully. Chromium checks against that container confirm that exhausted weekly/biweekly series are absent while remaining occurrences are visible, including a partial initial week.

The first full test run under `TZ=Asia/Jerusalem` had three solar-test failures; the same three reproduce on unchanged main. Those tests explicitly assume UTC. Existing Vite deprecation/chunk-size warnings remain. Native mobile and Obsidian runtimes were not exercised.


krelltunez (approved):
Verified the fix and tests locally, thanks!

PR #1725, opened 2026-09-19
Seat JOBO as a sixth desktop view behind an experimental flag

krelltunez:
Refs #1673 and #1623. The first slice of landing @Lcub3d's plan-versus-actual prototype in main incrementally, so that both sides build on one base instead of a fork that drifts.

**This adds no behaviour for a regular user.** A `joboEnabled` flag, off by default, and a view that only exists while it is on.

## What is in it

| Piece | Change |
| --- | --- |
| `constants/views.js` | `'jobo'` appended to `DESKTOP_VIEW_MODES`, key `6`, label key `sched.viewJoboShort` |
| keyboard, cycler, shortcut sheet | nothing — they read those maps |
| `App.jsx` | `joboEnabled`, persisted per device at `day-planner-jobo-enabled`, exposed on context |
| `SettingsModal` | an **Experimental** section with the switch; the per-device view switches only offer JOBO when it is on |
| `DesktopLayout` | renders `JoboView` when active; the summary strip stands down over it |
| `JoboView.jsx` | a placeholder, deliberately |
| locales | `sched.viewJoboShort`, `settings.experimental*`, `settings.enableJobo*`, a `jobo` namespace, in all eight |

## How "off" works

Off is expressed as **hidden**. `gateExperimentalViews` folds every experimental view whose flag is off into the hidden-views list `App` puts in context, and every consumer of that list already keeps a hidden view out of the cycler, the number keys, the shortcut sheet, the default-view picker and the stored-view fallbacks. So nothing that lists views had to learn the flag exists.

The stored list stays the user's own choices. The flag never writes into it, so turning it back on restores the view without a trip through Settings, and a device that turns the flag off does not sync a hidden JOBO to one that has it on.

`offeredViews` is the one other touch: the "views on this device" switches would otherwise show a JOBO toggle for a view the flag had already removed.

## Why the view is empty

What it shows comes after the ledger does. The prototype's ledger is localStorage-only, not carried through restore, reset or either sync tier, and `originalPlan` (a single field) already shipped once unable to survive a sync cycle. A view drawn from a store that cannot survive one is worse than no view, so the flag and the seat land first and the ledger (`src/jobo/`, following the Todoist integration's pure-core / persistence / hook split) is the next slice. None of it is here.

## Verification

Driven in the running app at 2000px, where the full cycler shows.

Flag **off** (default):

```
initial          View: MULTI (1/2/3/4/5 or C to switch)   joboOnScreen: false
after key 6      View: MULTI (1/2/3/4/5 or C to switch)   joboOnScreen: false
after 6× C       View: DAY                                 (five views, one full loop plus one)
stored 'jobo'    View: MULTI                               (falls back, stored hidden list untouched)
Settings         Experimental heading present, no JOBO switch under views on this device
```

Flag **on**:

```
initial          View: MULTI (1/2/3/4/5/6 or C to switch)
after key 6      View: JOBO  (1/2/3/4/5/6 or C to switch)  joboOnScreen: true
C from JOBO      View: MULTI                               (wraps)
Settings         JOBO in the default-view picker and in views on this device
```

Mutation-checked: a gate that never hides fails tests in three files (6 tests); an `offeredViews` that offers everything fails two. Full suite green at 4,630 tests, lint clean.

## What this makes possible

With this in main, the remaining slices are ordinary PRs against `main`: the ledger's pure rules into `src/jobo/core.js` (the piece @Lcub3d is best placed to take, and can start on now against these names), then its persistence and merge story with the field checklist in `CLAUDE.md` applied, then the view itself.

---
_Generated by [Claude Code](https://claude.ai/code/session_01M8k44pwgABpWA4bzd7RPm5)_


Lcub3d:
Looks good to me. I'll take the pure ledger/core slice in `src/jobo/core.js` from here, keeping it independent of persistence and React.

Issue #1726, opened 2026-09-19
[Feature] dayGLANCE Jobo: Landing the plan-versus-actual prototype in slices


krelltunez:
Tracking issue for landing @Lcub3d's Jobo prototype in main incrementally, behind the `joboEnabled` flag, so that everyone builds on one base. The prototype PR (#1673) is closed but remains the design reference: its thread holds the rationale for slicing rather than merging as one piece, and the full design document lives on @Lcub3d's branch at [docs/jobo-design.md](https://github.com/Lcub3d/dayGLANCE/blob/dayGLANCE-jobo/docs/jobo-design.md). The persistence design for the ledger is merged on main at [docs/jobo-ledger-persistence.md](https://github.com/krelltunez/dayGLANCE/blob/main/docs/jobo-ledger-persistence.md), reviewed over three rounds on #1744. The mobile follow-up (#1675) stays open and is held until the desktop shape settles, per slice 8. This issue tracks what has merged and what is next.

Ground rule for every slice: with the flag off, nothing changes for anyone. Each slice is an ordinary PR against `main`, reviewable on its own.

## Slices

- [x] **1. The seat.** `joboEnabled` flag (off by default, Experimental section in Settings), `'jobo'` registered as a sixth desktop view on key 6, placeholder view. #1725
- [ ] **2. The ledger's pure core.** `src/jobo/core.js` with tests and no storage or React behind it: the Do record shape, the status transitions (Completed, Partially Completed, Started, Not Started, and re-completing in a second session), and the timing and progress classification (Within Plan, Delayed, Overrun, Interrupted, Unplanned). Per the ledger design, core also owns the record constructor, the snapshot semantics (`title` and `planSnapshot` captured once, then preserved) and `pickJoboRecord`, the merge rule both sync tiers share. @Lcub3d is building this.
[...]

PR #1744, opened 2026-09-20
docs: JOBO ledger persistence and merge design (slice 3)

krelltunez:
Refs #1726 (slice 3), #1623, #1673. Design only; no code.

This is the slice that decides where Do records live, how they reach React state, how both sync tiers carry them, and what backup, restore and reset do with them. It is written before any code because it is the slice hardest to change later, and because two people are building against it: `src/jobo/core.js` (slice 2) is @Lcub3d's, the view is slice 5. The doc is `docs/jobo-ledger-persistence.md`.

**Revised after @Lcub3d's review below.** The collection design is unchanged; the record contract, the identity rule and the storage failure semantics were tightened. The doc marks each place a decision changed.

## The decision

**The ledger is a collection, not a task field and not a cache.** Each Do record is its own row with its own `updatedAt`, stored in IndexedDB, mirrored into React state by one hook that is the only writer, registered with the sync layer as one more `COLLECTION_KINDS` entry, and carried by every device whether or not it has JOBO on.

Two homes rejected, with reasons: on the task (unbounded per task, whole-entity LWW loses attempts, and every task field is another shot at the `normalizeField` resurrection hazard) and a single IndexedDB blob like the Todoist cache (derived data can merge as a blob; original user data cannot).

## Four invariants

1. A Do record never changes the plan.
2. **A record captures once, then preserves.** `title` and `planSnapshot` are core's, copied at creation and never refreshed from the live task by any later edit, persistence or remote apply.
3. **The flag gates the interface, never the data.** A device with JOBO off still loads, stores, pushes, pulls and merges. Otherwise its first sync pushes a payload without the collection.
4. Records reach React state, not only storage. The `originalPlan` lesson, in both directions.

## What changed in review

- **The record** is core's whole shape. `taskId` and `planSnapshot` may be null; an unavailable snapshot is not "unplanned". `endDate` is explicit. The snapshot is described honestly as the completion-time plan unless an earlier event supplies a better baseline.
- **Identity is ensure-present.** `do:${taskId}:${completedAt}` (every completion path stamps `completedAt`; checked), or template id plus instance date for recurring. Re-observation never rewrites or bumps `updatedAt`; the initial `updatedAt` is anchored to the source event, so any later edit or tombstone beats a late detector. Un-completing keys off the previous snapshot.
- **The ledger gets its own store**, `src/jobo/store.js`, over the shared IndexedDB plumbing rather than the cache helper's methods: that helper's `get()` falls through to the localStorage fallback on a failed read and its `set()` returns `false` rather than throwing. Strict read (absent is not failed, and failed never consults the fallback), checked write, and an atomic `update(fn)` in one readwrite transaction so two tabs see each other's appends. An unreadable ledger never becomes `[]`.
- **Applies during load are held** and merged after hydration.
- **Tombstones are never pruned.** The file tier's `syncHorizon` drops a stale local-only row past the window, but the vault tier is grow-only and would accept it. A kept tombstone loses nothing (about a hundred bytes each, deletions rare) and protects under every transport at any age.

## Testing

Eleven end-to-end scenarios, the five from review folded in, each walking the whole path rather than unit-testing a module, and the specific mutation that must fail each one.

The incidental finding from the first draft (IndexedDB names missing from the reset list) is fixed in #1745.


krelltunez:
@Lcub3d this is the slice 3 design: where Do records live, how they reach state, how both sync tiers carry them, and what backup, restore and reset do with them. No code yet, on purpose. I would like your eyes on it before it merges, since parts of it are a contract with the core you are writing.

The parts that touch your work directly:

- "The record" fixes six fields the sync layer depends on and treats everything else as yours. If any of those six is awkward for the core, now is the time.
- "Identity" gives detector-created records a deterministic id so two devices observing one completion collapse under last-writer-wins.
- The two open questions at the end, on whether `title` and `planSnapshot` belong in the core record shape or get added at write time.

Everything else, the four invariants and the storage choice especially, I would also welcome pushback on. Line comments on the PR are ideal; a reply here is fine too.

Lcub3d:
@krelltunez Thanks for writing this down before implementation. I agree with the main direction: a separate collection, per-record merge, no ledger array on the task, and the feature flag gating the interface rather than transport.

I worked through the document and relevant helpers with AI assistance. These are proposed contracts and review findings, not an implementation/test sign-off.

### 1. `title` and `planSnapshot` belong in the core record contract

My preference is to include both in the core shape. The caller supplies the title, copied plan values, ID and timestamps as ordinary inputs. Core constructs and validates the record without reading React state, storage or the clock.

The important rule is **capture once, then preserve**. Progress edits, interval corrections, persistence and remote applies should not refresh the snapshot from the live task. Renaming, rescheduling or deleting the task must not rewrite its historical records.

Please also allow `taskId: null` for an unlinked manual Do and `planSnapshot: null` when there genuinely was no timed plan. An unavailable historical snapshot is not necessarily the same as “unplanned”; we should not manufacture one from today's task.

One qualification to my earlier Final Plan description: a completion-only detector can capture the plan it observes at completion, but cannot prove what the plan was before work began if it changed meanwhile. Where an earlier start/recording event provides a reliable baseline, preserve it. Otherwise, I would document the completion-time baseline rather than promise history we did not capture.

The interval fields also need an explicit cross-midnight convention, such as an `endDate`.

### 2. Progress and reopening

I would keep the prototype's Started / Partially Completed / Mostly Completed / Completed vocabulary, with the serialized values defined once in core. Not Started is the derived condition of a plan without a recorded attempt, not a zero-duration Do.

For the transition discussed here:

- Completing creates a Completed attempt.
- Un-completing changes that associated attempt to Partially Completed, keeping its interval and snapshot.
- Completing in a second session appends another attempt; the earlier one remains intact.

This does not settle the separate JOBO-checkbox-to-native-task question from #1623.

### 3. Deterministic identity also needs idempotent creation

`do:${taskId}:${completedAt}` makes sense when both devices have the same stable source-event timestamp.

I would make the companion rules explicit:

- Preserve the previous completion key before an uncheck clears or replaces `completedAt`, so reopening targets the correct record. A later completion gets a new key; editing an interval never changes its ID.
- Re-observing a completion is ensure-present, not an edit with a fresh `updatedAt`. Otherwise a late detector can overwrite a user's correction or a tombstone despite having the right ID. Initial automatic versions should be anchored to the source event, not the later observation time.
- Missing completion timestamps and recurring occurrences need an explicit identity policy. Substituting each device's current time would defeat deduplication. Manual creation should also reuse its generated ID on retry.

### 4. Two persistence details to tighten

**Failure handling:** on this PR's base, `src/utils/idbKeyValue.js` already waits for transaction completion, but `set()` returns `false` if both storage paths fail, and `get()` can collapse a failed read into `undefined`.

So `await set(...)` alone is not a success check. The ledger needs a strict path distinguishing absent data from read failure, and must check write success before acknowledging a save or reloading after restore. An unreadable ledger must not become an authoritative `[]`. This could be an opt-in or ledger-specific API without changing the cache callers' behavior.

**Concurrent writers:** one hook is a good API boundary, but two tabs still have two hooks. Both can read `[A]`, then separately save `[A, B]` and `[A, C]`, losing one record before sync can merge it.

Keeping the array under one key is workable, but I would make read-current → apply the operation/merge by ID → write one atomic transaction, then refresh state from the committed result. Separate `get()`/`set()` calls or a per-tab queue do not provide that cross-tab guarantee. The localStorage fallback needs an explicit coordination or fail-closed policy too.

### 5. Lifecycle and tests

I agree with omitting the collection until loaded and awaiting restore. Incoming applies during loading should be held or merged after hydration rather than racing a stale initial read. Remote apply must preserve incoming timestamps. For the detector's remote-apply window, I read “defer” as holding the transition for the next quiet pass, not consuming it permanently.

Alongside the existing scenarios, I would add tests for:

- frozen title/plan after task changes and later Do edits;
- complete → reopen → re-complete, with a delayed duplicate after an edit or deletion;
- two tabs appending different records, and remote apply during initial load;
- failed reads/writes/restores without false success, reload or empty-ledger publication;
- unlinked/untimed Do and a cross-midnight interval.

Could we also document what happens when a stale device returns after the existing 60-day tombstone window? The fixed pruning window alone cannot guarantee that its old live copy will not reappear; we should identify the protection or explicitly record the limitation.

My proposed split is: **core owns the complete record shape, snapshot semantics and progress transitions; the hook/storage layer owns coordinated durable writes; sync merges those records without rebuilding them from today's task.**

The collection design can stay. These are the contracts I would tighten before building against it.

krelltunez:
@Lcub3d thank you. This is exactly the review the doc was asking for, and I have taken nearly all of it. The revised design is on the branch now; the changes and the reasoning:

**The record.** `title` and `planSnapshot` are core's, constructed from ordinary inputs with no state, storage or clock behind them. "Capture once, then preserve" is now invariant 2, with the second word doing the work: no progress edit, interval correction, persistence or remote apply ever refreshes them. `taskId: null` and `planSnapshot: null` are both allowed, and the doc says plainly that an unavailable snapshot is not the same as unplanned and the hook must never manufacture one from today's task. `endDate` is explicit. Your qualification on Final Plan is adopted as written: a completion-time snapshot is what a completion-only detector can honestly capture, an earlier event supplies a better baseline where one exists, and the doc promises no more than that.

**Progress.** The prototype's four values, serialized once in core. Complete creates a Completed attempt, un-complete drops that attempt to Partially completed keeping interval and snapshot, a later completion appends. Not started is derived, not a record. The JOBO checkbox question stays open on #1623 as you said.

**Identity.** You found a real hole. Re-observation is now ensure-present: it never rewrites a record or bumps `updatedAt`, and the initial `updatedAt` is anchored to the source event rather than the observation, so two devices produce byte-identical initial records and any later user edit or tombstone wins over a late detector. Un-completing computes the key from the previous snapshot, since the planner sees prev and next like `useCompletionLog` does. I checked the timestamp question against the code: every completion path stamps `completedAt`, so ordinary tasks always have a source-event key; recurring instances key on template id plus instance date, which is deterministic without `completedDatesTimestamps`, which is not present on every older instance. A key is never built from a device's current time. Manual records mint their id before the write so a retry reuses it.

**Storage.** I checked `idbKeyValue.js` and it is slightly worse than you described: `get()` falls through to the localStorage fallback when the IndexedDB read fails, so a broken read can come back as a stale or absent fallback value and look clean. The ledger therefore gets its own `src/jobo/store.js` over the shared plumbing: a strict read that distinguishes absent from failed and never consults the fallback on failure, a checked write the hook acknowledges only on success, and an atomic `update(fn)` that runs read, merge by id, put inside one readwrite transaction so two tabs see each other's appends. The localStorage fallback keeps read, merge, write synchronous in one turn, and its remaining cross-tab window is recorded as a limitation rather than hidden. An unreadable ledger leaves the hook unloaded and the payload without the key; it never becomes `[]`.

**Lifecycle.** Applies during load are held and merged after hydration. Remote apply preserves incoming timestamps. Defer means hold for the next quiet pass.

**Tombstones and the 60-day window.** I looked at what protects tasks today. The file tier has `syncHorizon`, which drops a local-only item older than `tombstonePrunedBefore` as a presumed zombie; the vault tier is grow-only and would accept a stale device's old live row. Rather than rely on one tier having a rule the other lacks, ledger tombstones are never pruned. A tombstone row is about a hundred bytes, deletions are rare, and a kept tombstone makes a returning stale copy lose on `updatedAt` under every transport at any age.

**Tests.** Your five are folded in, and the mutation list names what breaks each.

One thing I did not take, or rather took differently: the doc still says the hook is the only writer. Your split, core owns shape and transitions, storage owns coordinated writes, sync merges without rebuilding, is now the stated division at the top of "The record", and the single-writer rule is the hook's half of it.

The collection design is unchanged. If this reads right to you I will merge it and slice 3 can be built from it once `core.js` gives us the record constructor.

Lcub3d:
@krelltunez Thank you for working through the review so carefully. I have reread the revised document at `6a88a1f`.

The main direction now matches what I meant: core owns the complete record, snapshots are captured once and preserved, and the store handles strict reads, checked writes and atomic updates. I also agree with keeping the hook as the only public mutation entry point; that is compatible with the store coordinating writes across tabs. The retained tombstones and held applies address the previous lifecycle concerns.

I would keep those decisions. There are three remaining points I would clarify before treating the contract as settled—not a request to reopen the overall architecture.

### 1. Recurring occurrence identity versus attempt identity

The document uses `do:${templateId}:${instanceDate}` for recurring work, but also promises that completing again after reopening appends a new attempt.

For the same recurring occurrence, those rules conflict:

- Complete: create A under that key.
- Reopen: A becomes Partially completed.
- Complete again: the key is unchanged, so ensure-present does nothing and there is no second attempt.

The current recurring `toggleComplete` path already updates `completedDatesTimestamps[dateStr]` on each toggle. For new completions, could the key include that stable source completion stamp, with reopening using the previous snapshot's stamp?

Older occurrences without a stamp need an explicit legacy rule. Existing historical IDs should not be silently changed when a timestamp later becomes available, and missing attempts should not be reconstructed from the observing device's clock.

Please include the same recurring occurrence in the complete → reopen → re-complete test, not only ordinary tasks.

### 2. Equal source timestamps do not guarantee identical snapshots

Anchoring the initial `updatedAt` to the source event is right, but it does not by itself produce byte-identical initial records. A delayed device could observe the same completion stamp after the task's title or schedule changed, producing the same ID/version with different captured values.

We should either have both constructors consume the same persisted event snapshot, or define how differing initial candidates with the same ID/version converge, using the same deterministic rule in both sync tiers. The latter establishes convergence, not proof that the selected snapshot is the original pre-work plan.

A focused test would create two candidates with the same ID/version but different title/plan snapshots and exercise both merge orders through both transports. This is about initialization conflicts, not refreshing existing records from the live task.

### 3. The localStorage fallback cannot promise automatic recovery

I accept that the fallback lacks IndexedDB's transaction guarantee. However, the sentence saying a concurrent record is picked up on the next read overstates what merge-by-ID ensures.

If both tabs read `[A]`, then write `[A, B]` and `[A, C]`, B is absent from storage. If the tab holding B closes before another durable copy or sync exists, the next read cannot recover it. This can be data loss, not just temporarily stale visibility.

Please remove that recovery guarantee and state the degraded-mode policy explicitly. My preference for implementation is coordinated writes where supported, and otherwise a clearly indicated read-only/fail-closed mode. If writable fallback remains a deliberate trade-off, its possible loss should be stated plainly rather than assumed recoverable. The IndexedDB design itself does not need to change.

With these identity details and guarantees clarified, I support landing this as the slice-3 design baseline. Implementation and integration-test sign-off remain separate. The JOBO checkbox decision stays open on #1623, and the separate Life Planner prototype is not a prerequisite for this work.

As before, this is a design review with AI assistance, not a claim that the application implementation has been tested.

krelltunez:
@Lcub3d all three taken, and the second turned out to be bigger than you described. The revision is on the branch at 8f4586d.

**1. Recurring identity.** You are right, and I confirmed the mechanism you pointed to: both branches of the recurring `toggleComplete` write `completedDatesTimestamps[dateStr]` on every toggle, and so do `applySetCompletion` and the Obsidian bridge. The key is now `do:${templateId}:${instanceDate}:${stamp}`, with the stamp read from the snapshot in which the date is present. On an uncheck that is the previous snapshot, since the uncheck overwrites the stamp with its own time. A re-complete writes a new stamp and is a new attempt. Legacy rule: a completion observed with no stamp keys on template id and date alone, that id is permanent, a stamp appearing later is not a transition so nothing re-keys, and no attempt is ever reconstructed from the observing device's clock. Scenario 5 now runs for the same recurring occurrence, and once more with a delayed duplicate of the first completion arriving after the reopen.

**2. Convergence.** The byte-identical claim is gone. While checking what happens to two candidates with the same id and `updatedAt`, I found that both tiers resolve an exact timestamp tie as remote wins (`>=` in `dbAdapter.js` and in the file-tier merge). That is order-dependent: device A keeps B's copy, device B keeps A's, and they never converge. Tasks never hit it because two devices almost never share a `lastModified`; pristine ledger records share one by design. So the record gains `observedAt`, the detector's own clock at creation, and both tiers pick between copies with one shared function from core, `pickJoboRecord`: newer `updatedAt`, then lower `observedAt`, then smaller canonical JSON. Order-independent and idempotent, and the earlier observer is the one closest to the event. The doc says what you said: this establishes convergence, not proof that the chosen snapshot is the pre-work plan. "Local observer wins" was not available, because the hold in `useCompletionLog` mixes local and remote-applied edges on purpose. Scenario 4 gains the two-candidate case through both orders and both transports, with a mutation that lets each tier's own tie rule run.

**3. The fallback.** The recovery sentence was wrong for exactly the reason you gave, and it is removed. In its place: with Web Locks, which `useTodoistSync` already takes around its writes, `update(fn)` runs under a cross-tab exclusive lock named for the key, so the fallback has the same append guarantee as IndexedDB. Without Web Locks the ledger is read-only on that device. The store reports it, the hook refuses local mutations with a visible reason, and synced records are still merged into state so the device is not blind, only silent. That is the same fail-closed policy as the `writeLockUnavailable` error in the Todoist hook. Scenario 7 covers both fallback cases.

The collection design is unchanged. Unless you see something else, I will merge this as the slice 3 baseline and point #1726 at the main path. Implementation and integration-test sign-off stay separate, as you said.

Lcub3d:
@krelltunez Thanks — I reread 8f4586d. The three points from my follow-up are now addressed, and the core/store split looks clear.

One remaining detail: the file-tier example still passes horizon to mergeArrayById. That helper drops old local-only rows, which could include valid Do records or retained tombstones. Could we explicitly disable the task-style pruning horizon for joboRecords and add a regression covering both?

With that clarified, I’m comfortable with this as the slice-3 design baseline. I’ll provide the record constructor and pickJoboRecord in pure core, keeping storage, React wiring and completion detection outside it. Implementation and integration-test sign-off remain separate.

krelltunez:
@Lcub3d good catch, and a real one. I read `mergeArrayById` in the sync package: a local-only row older than the remote's `tombstonePrunedBefore` is dropped as a presumed zombie. That compensates for the task tombstone map being pruned at sixty days. The ledger keeps its tombstones as rows forever, so it needs no such rule, and with the horizon in place an offline device, or a merge against a payload from a build that predates the collection, would have dropped every Do record and every tombstone older than sixty days. The merge now passes `null`, as `areas` already does at the same call site. Scenario 3 asserts that an old local-only record and an old local-only tombstone both survive, and that the tombstone still beats a stale live copy arriving afterwards; passing the horizon is the mutation that breaks it. Revision at e3d4f5d.

Merging this as the slice 3 baseline. Thank you for the record constructor and `pickJoboRecord` in pure core; storage, React wiring and the detector stay outside it, as agreed.

PR #1762, opened 2026-09-21
feat(jobo): the ledger's persistence and merge (slice 3)

krelltunez:
Refs #1726 (slice 3), #1744, #1788. Ready for review.

Implements `docs/jobo-ledger-persistence.md` in full, against the record shape slice 2 (#1788) landed: where Do records live, how they reach React state, how both sync tiers carry them, and what backup, restore and reset do with them. Nothing draws anything; the JOBO view stays the slice 1 placeholder. With the flag off nothing changes for anyone, and with it on nothing changes either yet, since no slice 4 detector writes records.

## Storage and state

**`src/jobo/store.js`**, the ledger's own door to IndexedDB, over the shared open-database plumbing. A strict `read()` that distinguishes absent from failed and never consults the localStorage fallback on failure; a checked `write()`; an atomic `update(fn)` that runs read, merge, put inside one readwrite transaction so two tabs appending together both land. The fallback takes a Web Lock, the pattern `useTodoistSync` uses, and is read-only where Web Locks is missing too.

**`src/jobo/ledger.js`**, the state machine without React: records are `undefined` until a strict read succeeds and an unreadable ledger is never published as empty; remote applies before the load are held and merged after hydration; every mutation goes through `update(fn)` and state is the committed value; remote applies keep incoming timestamps; a read-only device refuses local mutations but still merges what sync brings. The merge rule is core's `pickJoboRecord`, the one function the hook and both tiers share. `mergeJoboCollections` is the file tier's merge with the change-flag contract, and `restore` is the checked replace the restore paths await.

**`src/hooks/useJoboLedger.js`**, the thin subscription and the only writer of `joboRecords`.

## Both sync tiers

**Vault (`src/sync/dbAdapter.js`).** `joboRecords` joins `COLLECTION_KINDS` on `updatedAt`, and is insert-only so every pulled copy runs through `applyRemoteJobo`, which resolves two copies of one record with `pickJoboRecord` (newer `updatedAt`, then lower `observedAt`, then canonical JSON). The engine's own tie rule is "remote wins", which depends on which side is remote: two devices holding pristine copies of one record, same id and same source-anchored `updatedAt` but different snapshots, would each keep the other's and never converge. A local win returns the row for re-push, the way bundle supersets do; an identical pulled copy is a no-op. A row-level delete is ignored: a ledger deletion is a `deleted: true` row.

**File (`src/mergeSync.js`).** The upstream merge drops unknown keys, so the collection is re-merged by id with the same rule. `localChanged` only when the merge differs from this side, `remoteChanged` only when it differs from the remote, neither for equal copies. No sync horizon is passed: it drops old local-only rows, which for the ledger are real records and the kept tombstones, and a merge against an older build's payload would otherwise drop every row older than sixty days. A side without the key did not carry the collection and is never read as empty.

## App

`buildSyncPayload` carries the collection only once loaded (`undefined` means not loaded, which both tiers treat as "does not carry it"; `[]` would claim empty). `applyEngineData` hands pulled rows to the hook. Both backup payloads include the ledger. Every restore path, file, folder, local auto-backup and remote, awaits the checked write before reloading and says so if it fails. `KNOWN_INDEXEDDB_NAMES` gains `dayglance-jobo`, which the reset scan from #1745 demands. The context exposes `joboRecords`, the load and writable state, and `recordJobo` for slices 4 and 5. `CLAUDE.md` carries the load-bearing rules.

## Tests

The store on both paths (`fake-indexeddb` and the fallback with and without locks), the controller lifecycle, and `src/sync/joboRecordsSync.test.js`, which walks the design doc's scenarios end to end on both tiers: save to state to payload to vault row to the other device's state; the key omitted while unloaded; an older build's payload merging against one with the collection; an old local-only record and tombstone surviving past the tombstone window; two devices converging on the earlier observer in both orders through a shared vault (`dbVaultSim`) and through the file merge; a user edit and a tombstone beating late re-observations; an identical copy raising no flag; a row delete leaving the ledger untouched; restore through a fresh controller, and a failed restore write not acknowledged. Fourteen guards mutated one at a time across the two halves, each confirmed to fail its test. The losslessness gate learns the new insert-only kind.


krelltunez:
@Lcub3d - once #1762 lands I will take slice 4, the detector, since it sits in the same sync and hook code. Slice 5, the view, is yours whenever you want to start it, and it can begin against the #1762 branch now since the context already exposes `joboRecords` and `recordJobo`.

Lcub3d:
@krelltunez I mostly looked at how Slice 2 feeds into Slice 3, and that handoff looks solid to me. #1762 seems to fit the merged #1788 shape cleanly.

While reading through the persistence lifecycle, I did notice a few cases I’m not completely sure about. I’d appreciate your recheck on whether these are real issues:

1. `applyEngineData()` does not await `applyRemoteJobo()`, while the DB sync cycle can save its snapshot and advance the cursor immediately afterwards. My concern is that a failed IndexedDB write could be acknowledged by sync before the Do is actually durable.

2. The restore callers appear to ignore `{ ok: false }` from `restoreJoboOrWarn()` and continue applying the backup / reloading. If I’m reading that correctly, the restored Plan state and the existing Do ledger could end up representing different points in time.

3. In `ledger.load()`, if flushing held remote records fails, it looks like the merged-but-uncommitted value can still be published, the held batch cleared, and the error then cleared. I would have expected state to expose only the committed value and the failed batch to remain retryable.

One smaller concurrency case I noticed: if another apply arrives while the first held batch is awaiting `store.update()`, it looks possible for that second batch to remain held after `loaded` becomes true. Would it be safer to drain the held queue until it is stable?

And on Slice 5, I’m really glad to see the Jobo/ledger idea taking shape as part of dayGLANCE. For the visual direction, I’m happy to follow your lead. I can either put together a first pass for us to react to, or work from a few design principles you’d like the view to follow first. Either works for me.

krelltunez:
@Lcub3d thank you, that was a precise read. Three of the four were real, and the fourth was real in a way the branch had not written down.

1. **Cursor before durability.** You are right about the ordering, and awaiting the write would not change it: the engine advances the pull cursor on the mirror write, and the ledger write is a hop later. What makes it safe is the engine's snapshot-delete guard. The committed mirror is the snapshot, the ledger state that feeds the payload lacks the row, and a row that vanishes from the payload with no tombstone is a suspected glitch, never a delete: it is re-fetched by id and re-committed every cycle until it lands, and the snapshot is withheld until then. The other device is untouched throughout. That was true before but untested; scenario 12 now runs it through the real engine with a store that fails the first write, and the test fails if the heal is unavailable or the kind is released from the baseline.
2. **Restore.** Real; the alert fired and the restore went on. Every restore path now writes the ledger first, awaited, and aborts on failure. First because it is the one step that can fail, and the localStorage replacement after it cannot.
3. **Failed flush.** Real, and fixed the way you describe: state shows only what the read returned, the error is reported, and the batch stays held for the next apply or load to retry. A loaded apply whose write fails is held the same way, and the queue merges by id so a re-delivered row never grows it.
4. **Mid-flight apply.** Real; the flush now drains until stable.

All in fabaa91, with a test and a mutation check for each. On slice 5: a first pass to react to, please. The one principle I would ask for up front is that the view reads `joboRecords` and writes through `recordJobo` only, so the view never has a second opinion about what is on disk.

Issue #1686, opened 2026-09-16
[Feature] dayGLANCE Jobo: Connect long-term direction to annual goals, monthly priorities and tasks


krelltunez:
Split out of @Lcub3d's roadmap comments on #1623. This combines two related ideas: the long-term hierarchy, and making goal-to-task relationships more visually understandable.

## The idea

A chain from life direction down to scheduled time:

**Life direction → Annual goals → Monthly priorities → Concrete tasks → Time allocation → Actual effort and review**

And the same chain read in reverse: daily records should let you ask whether what you actually spent time on matched the direction you said you wanted, and whether the goals themselves need adjusting.

The first interface could be modest: a manually filled page with light guidance. When setting annual goals, a template prompts you to think about which directions matter this year and how many goals genuinely deserve attention. The relevant life wish appears as a short note under the annual goal, so the answer to **"why do I want this?"** stays visible.

Collections like reading or film lists could link to goals where useful, without forcing every interest to become a measurable objective. That restraint seems important to the idea rather than incidental to it.

@Lcub3d also raised making goal decomposition more visually explicit, possibly canvas-like, and noted it overlaps with lifeGLANCE territory while possibly wanting a different approach.

## Their framing, worth keeping

> If life were a Go board, I imagine the horizontal axis as all the directions in which we choose to make an effort, and the vertical axis as time. I'd like software that is not mainly about storing static information like birthdays or phone numbers, but about helping us decide more clearly where to place the next stone at the point where direction and time intersect.

## If the decomposition view gets built, it should reuse the existing visual language

The dayGLANCE logo is a single point inside a calendar frame, fanning into several diverging paths that run left to right. That was deliberate: one task starts as one thing and ends as many.

@Lcub3d noticed the same shape reappearing in the Jobo prototype, where the DAY view draws curves from one Plan block to each actual interval recorded against it. They arrived at it independently rather than copying it, which is a reasonable sign the shape genuinely fits the idea rather than being decoration.

A goal decomposition view is the same structure a third time: one direction branching into priorities branching into tasks, one becoming many. So if this is built, it should extend that language rather than introduce a second unrelated one.

## Why this is the hardest one

This touches the information architecture directly. dayGLANCE already has Goals, Projects, Areas and GTD Frames; this proposes a layer above them and a stronger spine through them. That is not an additive feature, it is a restructuring, and @Lcub3d explicitly left it out of the current prototype for that reason.

[...]

PR #1764, opened 2026-09-21
fix: prevent desktop header dates from overlapping weather and actions

Lcub3d:
## Summary

Standalone Header follow-up to #1686, based on main (67ac6e0). No Life Planner, chooser, experimental entry points, or Jobo data changes.

The absolutely positioned date group can overlap weather/actions at constrained widths. This fix reserves grid tracks for all three groups, lets the date label truncate with a full-text tooltip, and keeps navigation buttons from shrinking. The 80px height, month popup, Day Dial, weather breakpoints and action callbacks are preserved.

Only DesktopHeader.jsx and DesktopHeader.test.jsx change.

## Validation

Fork CI passed: production dependency audit, lint, all 4,984 tests across 324 files (including 8 Header tests), and production web build:
https://github.com/Lcub3d/dayGLANCE/actions/runs/35631402818

The CI-only branch adds a workflow on top of this exact fix; the workflow is not included in this PR.

264 isolated Chromium layout cases across 721-2560px passed; the old layout overlapped in 108. These used source-derived fixtures with mocked contexts/icons, not a full application or Windows/Electron run.

Manual review: weather on/off, optional action buttons, long localized dates, previous/next/Today, month picker and Day Dial.

AI-assisted implementation and test preparation. Related to #1686; does not close it.


krelltunez (changes requested):
Built main and this branch side by side and drove both with mocked weather. The overlap is real: with the Obsidian button enabled and a non-today date selected, main at 780px buries the Today button under the calendar sync button and at 850px they touch. This branch keeps every group clear from 780px to 2000px in English and German, in MULTI and WEEK. Your tests pass here, lint is clean, and the Tailwind compile test is a good guard for an arbitrary-value class.

**One change before merging.** Dropping `min-w-[13rem]` on the date button means the previous and next chevrons now move as the label changes length while paging, which was the reason the fixed width existed. Give the label a basis of that width and let it shrink only under pressure: replace `min-w-0 truncate` with `basis-52 shrink min-w-0 truncate`. That keeps the chevrons still at ordinary widths and still yields in the narrow case your fix targets.

One note, not a blocker: with both side tracks at `minmax(max-content, 1fr)` the date drifts off centre once one side outgrows an equal share, which only happens above 2300px when the daily-content card appears. Fine as is.

Thank you for carving this out of the larger branch and keeping it to two files. That is exactly the shape that makes it easy to take.

Lcub3d:
@krelltunez Addressed in d3f4ed4. I added the requested shrinkable basis, paired with w-52: browser checks showed that basis alone still let short labels contract the auto grid track. The preferred width keeps the chevrons stable while still allowing the label to shrink at narrow widths.

Lint, all 4,985 tests, the production web build, and 384 rendered-component layout cases passed. The PR still changes only the same two files, and the accepted ultrawide centering behavior is unchanged. Ready for another look, thanks!

krelltunez (approved):
Rebuilt the new head and measured it. Paging across a month boundary at 1230px and 1400px, the label holds 13rem and both chevrons sit at exactly the same pixel as on main. At 780px in German the label yields to 160px, the chevrons still do not move, and the Today button is clear. The w-52 pairing is right for the reason you give: the auto track takes its size from content, not from flex-basis. Lint, tests and the production build pass here. Approving; thank you for the quick turnaround.

Issue #1763, opened 2026-09-21
[i18n] Some UI text ignores the app language (hard-coded English, browser-locale dates)


polymatheiia:
While adding Ukrainian and Polish translations (see #929) I hit a number of places that a translation file can't fix. They affect every non-English language, not just those two. Everything below is against current `main`.

### 1. Hard-coded English text that never goes through `t()`

Several of these hand-pluralize with an `'s'` suffix, which can't work in languages whose nouns change by number (Polish, Ukrainian, Russian, …).

- [x] `src/App.jsx:2295`: "GLANCEvault: N items couldn't be read". `sync.form.vaultSkipped` already exists and isn't used here.
- [x] `src/App.jsx:4674` (and around it): iCal import toast, "Imported N events", "No events found in the file", "Removed N file-imported events".
- [x] `src/App.jsx:8165` and `:8261`: "Tasks Scheduled" / "N tasks moved to future slots".
- [x] `src/App.jsx` around 5648–5670: the calendar-sync notification.
- [x] `src/components/SpotlightModal.jsx:154`: "N results".
- [x] `src/components/CalendarHeader.jsx:422`: "N deadlines".
- [x] `src/utils/obsidianBinRestore.js:86`: "N tasks were restored from the recycle bin…".
- [x] `src/components/DeadlinePickerPopover.jsx:83`: hard-coded English month abbreviations in the picker header ("Sep 2026" in every language).

### 2. Dates and times use the browser locale (or `en-US`), not the app language

- [x] `toLocaleDateString(undefined, …)` in `sched/SchedView.jsx`, `sched/SchedTaskCard.jsx`, `sched/SchedDashboard.jsx`, `projects/ProjectPlanner.jsx`, `projects/HyperGlanceEditor.jsx`. Elsewhere `formatLocalizedDate(…, language)` already does this correctly.
- [x] `src/utils/bridgeStatus.js:61`: `describeAgo(then, now, locale = undefined)` is called without a locale, so the Bridge "last seen" text follows the browser language.
- [x] `src/trmnl.js:196-197`: the e-ink display's weekday and date are always `en-US`.

### 3. Durations built as `1h 30m`

[...]

PR #1781, opened 2026-09-22
Add Ukrainian and Polish translations

polymatheiia:
## Summary

Adds Ukrainian (`uk`) and Polish (`pl`) locale bundles, translated from `en`
(machine-assisted, then hand-checked by a native speaker of both languages
against actual usage in code: plural keys whose count is always 1, fragments
embedded in other strings, second-person address).

- `public/locales/uk` and `public/locales/pl`, full key parity with `en`,
  picked up by the existing directory glob — no change to `locales.js`.
- Fixes the weekly review's date range, which was hand-built in English word
  order for every language (`вер. 15 — 21, 2026`). It now goes through
  `formatDateRange`, so this also fixes German (`Sep 15 — 21, 2026`) and
  Chinese (`9月 15 — 21, 2026`), not just Ukrainian/Polish.
- Reworks the plural-coverage test that previously scoped itself to
  `it.each(['pl', 'uk'])`. That scope silently stopped guarding
  de/es/fr/it/pt-BR/pt-PT for the plural forms they do need — mutation-tested
  by deleting `sync.form.vaultSkipped_one` from `de`, which passed the old
  test and fails the new one. The new version derives the categories a
  language's own `Intl.PluralRules` selects for counts the app can produce
  (0–1000) instead of a hardcoded exemption list, so es/fr/it/pt's `many`
  (reserved for millions) drops out on its own.

**Update:** rebased onto current `main` to pick up #1765 (Italian ordinal
`many` forms for 8th/11th), which the reworked test above also covers. That
resolved the one previously-known failing test. Rebasing also surfaced two
new English keys `main` gained since this branch was cut
(`settings.location`, `settings.locationHelp`, the Location settings block) —
translated both to restore parity.

## Not included

Android/iOS native strings. Some UI text bypasses i18n entirely
(hard-coded English toasts, browser-locale dates, hand-rolled h/m duration
formatters) — tracked in #1763, being addressed there section by section
(see #1782 for section 2).

## Test plan

- [x] Full suite: 5061 tests, only the 3 pre-existing `solar.test.js`
      failures (reproduced identically on clean `main`, date/timezone-
      dependent, unrelated)
- [x] `src/locales.test.js`: 114/114 passing
- [x] `eslint` clean on the changed source file
- [x] Key parity across all ten locales, no English value changed, no key
      removed


krelltunez:
Merging this first. Validated on Linux against current `main`: full suite 5061/5061, lint clean, key parity exact across all ten locales.

The guard rework is right, and I mutation-tested it rather than taking the description on faith. All five cases fail as they should:

```
de loses vaultSkipped_one   caught   (the regression the old scope let through)
pl loses a _few form        caught
uk loses a _many form       caught
it loses ordinal _many      caught   (guards #1765 going forward)
pl loses a plain key        caught
```

Deriving the categories from `Intl.PluralRules` over a real count range, rather than a hardcoded exemption list, is better than what I suggested: it means the next language added never needs a scope edit, and es/fr/it/pt's millions-only `many` drops out on its own.

Also noting for the record that your `solar.test.js` failures don't reproduce here. Clean `main` is 5002/5002 in my container, so those are timezone-dependent on your end as you suspected, not something you introduced.

Order for the other two: #1782 next, then #1783. Details in those threads.

PR #1782, opened 2026-09-22
i18n: route dates and speech recognition through the app language

polymatheiia:
## Summary

Section 2 (plus §4, per your note that it could ride along) of #1763.

- **Dates followed the browser, not the app language.** `toLocaleDateString(undefined, ...)` in `sched/SchedView.jsx`, `sched/SchedTaskCard.jsx`, `sched/SchedDashboard.jsx`, `projects/ProjectPlanner.jsx`, `projects/HyperGlanceEditor.jsx` — routed through the existing `formatLocalizedDate(date, options)` instead, matching the pattern already used everywhere else.
- **`describeAgo()` was called without its locale param.** `bridgeStatus.js:61` already accepts one; `BridgeStatusPanel.jsx` and `BridgePairingPanel.jsx` just never passed it, so the Bridge "last seen" text followed the browser language.
- **Voice input (§4): `recognition.lang = navigator.language`.** A Ukrainian UI on an English-configured browser recognized English speech. `recognition.lang` now comes from a new `speechRecognitionLocale()` helper in `localeFormatting.js`:
  - **Bare app tag** (`en`, `uk`): the browser's tag refines it when both name the same language, so regional speakers keep their tag (`en` + `en-GB` → `en-GB`, `de` + `de-AT` → `de-AT`). A different browser language is ignored (`uk` + `en-US` → `uk-UA`).
  - **App tag with its own region** (`pt-BR`, `zh-CN`): an explicit choice, kept as is.
  - **Otherwise**, the app tag is maximized to add a region (`uk` → `uk-UA`), dropping the script subtag that `Intl.Locale.maximize()` also adds (`uk-Cyrl-UA`), which the Web Speech API doesn't expect. An unparsable browser tag falls through to this path.
  - Covered by a table for all ten shipped locales and a table of app/browser mismatch pairs.
- **`App.jsx:2295`'s GLANCEvault-skipped toast was hardcoded English** with manual `'s'` pluralization (`couldn't be read`). `sync.form.vaultSkipped` already exists with correct plural forms in every shipped locale — wired it in, as you flagged it could ride with this PR.

No new locale keys, so no translation burden.

## Test plan

- [x] Full suite: 5079 passing. The only failures are three in `solar.test.js`, which is timezone-dependent and fails identically on clean `main`.
- [x] `speechRecognitionLocale` tests:
  - all ten shipped locales, with no browser tag;
  - app/browser mismatch pairs: regional refinement, a different browser language, an explicit app region, a browser tag with no region, an unparsable tag, and no tag.
- [x] Mutation-checked each guard: dropping the bare-tag check, dropping the same-language check, or rethrowing from the parse `catch` each fails a test.
- [x] `eslint` clean on every changed file.


krelltunez:
The date routing and the `describeAgo` locale threading are both right, and I confirmed `formatLocalizedDate` already defaults its third argument to `activeLocale()`, so omitting it is correct rather than accidental. Full suite 5012/5012 here, lint clean.

One thing to fix before I merge.

`speechRecognitionLocale()` derives the tag from the app language alone, so a bare `en` maximizes to `en-US`. That regresses regional speakers who previously got their own tag from `navigator.language`:

```
en-GB speaker, English UI   en-GB  ->  en-US
de-AT speaker, German UI    de-AT  ->  de-DE
es-MX speaker, Spanish UI   es-MX  ->  es-ES
fr-CA speaker, French UI    fr-CA  ->  fr-FR
```

Recognition accuracy is region-sensitive, so for accented English in particular that's a real downgrade, and it lands on far more users than the case the PR fixes.

The fix keeps your fix intact. Let the browser refine the tag only when the app tag names no region of its own, since a bare tag expresses no regional preference while `pt-BR` or `zh-CN` is an explicit choice:

```js
export const speechRecognitionLocale = (language = activeLocale(), reported = navigator?.language) => {
  const app = new Intl.Locale(language);
  // A bare app tag ("en", "uk") expresses no regional preference, so the
  // browser's own tag refines it where the two agree on the language. A tag
  // that already names a region ("pt-BR", "zh-CN") is an explicit choice.
  if (!app.region && reported) {
    try {
      const nav = new Intl.Locale(reported);
      if (nav.language === app.language && nav.region) return `${nav.language}-${nav.region}`;
    } catch { /* unparsable navigator tag: fall through */ }
  }
  const max = app.maximize();
  return max.region ? `${max.language}-${max.region}` : max.language;
};
```

I ran that across twelve cases and they all behave: `en`+`en-GB` gives `en-GB`, `uk`+`en-US` still gives `uk-UA` (your bug, still fixed), `pt-BR`+`pt-PT` stays `pt-BR`, and an unparsable navigator tag falls through instead of throwing. Worth extending your locale table to cover a couple of the app/browser mismatch pairs, since that is the interesting axis rather than the tag shape alone.

Will merge this next.

polymatheiia:
Thanks. You're right that it regressed regional speakers. I've applied your version, rebased onto current `main`.

One small change: the default reads `globalThis.navigator?.language` rather than `navigator?.language`, since a bare identifier throws a ReferenceError before the `?.` is reached anywhere `navigator` isn't declared.

The tests now cover the app/browser axis:

| App | Browser | Result |
|---|---|---|
| `en` / `de` / `es` / `fr` | `en-GB` / `de-AT` / `es-MX` / `fr-CA` | browser's tag |
| `uk` | `en-US` | `uk-UA` (the original bug, still fixed) |
| `pl` | `de-DE` | `pl-PL` |
| `pt-BR` | `pt-PT` | `pt-BR` |
| `zh-CN` | `zh-TW` | `zh-CN` |
| `en` | `en` (no region), unparsable, none | `en-US` |

The existing single-argument table now passes `null` for the browser tag, so it no longer depends on the test environment's `navigator.language`.

Mutation-checked each guard:
- dropping `!app.region` fails 2;
- dropping the same-language check fails 2;
- rethrowing from the `catch` fails 1.

Full suite 5079 passing. The only failures are the three in `solar.test.js`, which also fail on clean `main` here (timezone-dependent). Lint is clean.

krelltunez:
Merging. 5082/5082 against current `main`, lint clean.

The refinement rule is right, and `globalThis.navigator?.language` is a genuine improvement on what I proposed rather than a style preference: a bare `navigator?.language` throws a ReferenceError before the optional chain is reached, since `?.` guards null and undefined values, not undeclared identifiers. That would have taken out voice input in any non-browser context that loads the module.

I mutation-checked the three guards and they fail exactly as you describe: dropping `!app.region` fails 2, dropping the same-language check fails 2, rethrowing from the `catch` fails 1. Passing `null` explicitly in the single-argument table is the right call too, since a test that reads the runner's own `navigator.language` passes or fails by accident.

Thanks for catching the regression rather than just applying the patch.

PR #1783, opened 2026-09-22
i18n: consolidate duration formatting into formatDuration(minutes, t)

polymatheiia:
## Summary

Issue #1763 section 3, partly: see **Scope** below. Eight hand-rolled `h`/`m` formatters — in `HyperGlanceModeModal`, `MobileListView` (×2), `MobileSettingsPanel`, `MobileBottomSheets` (×6 inline), `WeeklyReviewModal`, and `MonthStats` — duplicated the same minutes → "Xh Ym" logic behind three overlapping key families (`common.hoursShort`, `focus.hoursMinutesShort`, `weeklyReview.durationHoursMinutes`). Replaced all of them with one exported `formatDuration(minutes, t)` in `src/utils/formatDuration.js`, with its own tests.

## What moved where

- New home for the canonical triplet: `common.durationMinutes` / `durationHours` / `durationHoursMinutes`, relocated from `weeklyReview` (which only had two callers — `WeeklyReviewModal` and `MonthStats` — so `common` is the right home now that it's used app-wide).
- `common.hoursShort` and `focus.hoursMinutesShort` are now unused and removed from every locale bundle.
- `common.minutesShort` and `voice.minutesShort` are **untouched** — deliberately out of scope. They're a separate family of ~40 call sites (duration pickers, task chips elsewhere) that don't do hour math at all.
- `daySummary.js`'s own `formatMinutes` (used by `TitlebarSummaryStrip`/`SummaryStrip`) is also untouched — it's a real, pre-existing i18n gap (hardcoded English "h"/"m", never localized) but a different bug than the one this section asked me to fix.

## One visible behavior change worth a look

`MobileListView`'s `durLabel`/`countdownText` previously fell back to `common.minutesShort` ("45 min") for the under-an-hour case. They now go through `formatDuration` → `common.durationMinutes` ("45m") for consistency with every other call site this PR touches. This means the mobile list view's compact duration/countdown chips will read "45m" instead of "45 min", while `MobileTimeGrid`/`InboxSidebar`/`TimelineTaskCardContent` (untouched, out of scope) still read "45 min" for their own duration chips. Happy to revert just that piece and keep `common.minutesShort` there if you'd rather the two mobile views stay visually identical to the desktop equivalents for now.

Also: `HyperGlanceModeModal`'s elapsed-timer label now renders "2h" instead of "2h 0m" on an exact hour (matches the "no zero-valued part" rule the other five call sites already had).

## Scope

This covers eight formatters, but not every site section 3 names. Several call sites still hand-roll their own after this PR (`RecurringSeriesRow`, `SchedTaskCard`, `WeekView`, `GlanceSidebar`, `MobileGlanceSection`, `HyperGlanceBar`, `App.jsx`, `useNewTaskInput`, `trmnl.js`, `quickAddParser`; the full list is in the #1763 status comment). Moving them onto `formatDuration(minutes, t)` is a mechanical follow-up and is not part of this PR.

## uk/pl

#1781 has merged, so this branch is rebased on current `main` and carries its own uk/pl migration: `weeklyReview.duration*` moves to `common.duration*`, and `common.hoursShort` / `focus.hoursMinutesShort` are dropped. The values are unchanged. The PR has no merge-order dependency on #1782 or #1784.

## Test plan
- [x] `formatDuration.test.js` — minutes-only / hours-only / both, fractional-minute rounding, negative/NaN clamping to 0
- [x] `npx vitest run`: 5064 passing. The only failures are three in `solar.test.js`, which is timezone-dependent and fails on clean `main` too.
- [x] `locales.test.js`: without the uk/pl migration, the four coverage checks fail (pl/uk "covers every key in en" / "carries no keys that en does not"). With it, they pass.
- [x] `eslint` on all touched files
- [x] Repo-wide grep confirms no remaining references to the removed keys


krelltunez:
Good consolidation, and thank you for the scoping notes. Naming `daySummary.js`'s `formatMinutes` as a real pre-existing gap you deliberately left alone is exactly right: it is a separate bug (German users see a hardcoded `2h 15m`) and folding it in here would have widened the PR. Same for `common.minutesShort`. Full suite 5008/5008 here, lint clean.

## Merge order, since you asked

**#1781, then #1782, then this one.** #1781 and #1782 compose cleanly on their own, 5071/5071, so they go in first and neither needs anything from you.

Your conflict prediction was right, and I measured it: #1781 plus this PR fails 4 tests in *either* order, so the order genuinely does not matter. There is no git conflict, `WeeklyReviewModal.jsx` auto-merges fine. It is purely the locale keys, and the fixup is five keys in each of `pl` and `uk`:

- rename `weeklyReview.durationMinutes` / `durationHours` / `durationHoursMinutes` to `common.duration*`
- delete `common.hoursShort` and `focus.hoursMinutesShort`

Your existing values carry over untouched (`{{hours}} godz. {{minutes}} min`, `{{hours}} г {{minutes}} хв`). I applied exactly that and ran all three PRs together: 5077/5077 green. So rebase this on `main` once the first two land, fold the five-key migration in, and I will merge.

## On the "45 min" to "45m" change

Taking it as written, no need to revert. Making the ladder internally consistent within a component is worth more than matching the sibling views, and the remaining `minutesShort` sites are a coherent follow-up rather than a reason to hold this. If you want that follow-up it would be welcome, but it is not a condition of merging.

`HyperGlanceModeModal` rendering `2h` instead of `2h 0m` is a straightforward improvement.

krelltunez:
Correction to my last comment: I said to rebase once the first two land. Only #1781 was ever a dependency, and it is merged now, so you are not waiting on #1782. This PR and #1782 share no files and merge cleanly against each other in either order.

The red checks above are the current state rather than a regression you introduced. The earlier green run happened about twenty minutes before #1781 merged, so it validated this branch against a `main` that had no uk/pl bundles at all. Now that they exist, the four failures are `pl` and `uk` against `covers every key in en` and `carries no keys that en does not`.

That is the uk/pl migration you offered to push once you knew the order. Rebase onto current `main` and fold it in: add `common.durationMinutes` / `durationHours` / `durationHoursMinutes` to both bundles, drop `common.hoursShort`, `focus.hoursMinutesShort` and the three `weeklyReview.duration*`. Your existing values carry over unchanged, so it is a rename rather than new translation work. That clears all four.

polymatheiia:
Rebased onto current `main`, with the uk/pl migration folded in as you described. In both bundles:
- `weeklyReview.durationMinutes` / `durationHours` / `durationHoursMinutes` are now `common.duration*`, with values unchanged;
- `common.hoursShort` and `focus.hoursMinutesShort` are removed.

I checked the four `locales.test.js` failures fail without that commit and pass with it. Full suite 5064 passing. The only failures are the three in `solar.test.js`, which also fail on clean `main` here. Lint is clean.

Thanks for the section 3 audit on #1763. I've added a Scope section to the description so the PR doesn't read as covering the whole section, and I'll take the remaining call sites as a separate follow-up onto `formatDuration(minutes, t)`.

krelltunez:
Merging. 5067/5067 against current `main`, lint clean, and the uk/pl migration lands the four coverage checks green.

Adding the Scope section is the right resolution. The abstraction here is the part that matters and it is correct; the rest of the section is a mechanical follow-up onto `formatDuration(minutes, t)` rather than anything this PR needed to carry. I will leave #1763 open for it.

One thing I checked before merging, since it is the shape of problem that bit this PR the first time round: this and #1784 both edit the uk/pl bundles, so I stacked them. They compose cleanly in either order, 5093/5093, because your migration removes the stale duration keys while #1784 only adds new ones. No third follow-up needed.

Taking the `45 min` to `45m` change as written, per my earlier comment.

PR #1784, opened 2026-09-22
i18n: translate hard-coded English text (issue #1763, section 1)

polymatheiia:
## Summary

Issue #1763, section 1 ("Hard-coded English text that never goes through `t()`"). This covers every checkbox except `App.jsx:2295` (the `vaultSkipped` toast), which #1782 already fixes. It isn't duplicated here, so the two PRs don't make competing changes to the same line.

- **iCal import toast** (`App.jsx`, import/empty/remove): the `${count} event${count !== 1 ? 's' : ''}` pattern is replaced with i18next count keys. The toast title has its own key, `sync.icalImportTitle`. `settings.importIcs` stays with the menu item, where de/fr/es/pt use an imperative that reads wrong as a toast title.
- **Tasks Scheduled / Tasks Rescheduled toasts** (`App.jsx`): the same pluralization fix, with new `app.tasksScheduledTitle` / `tasksPlacedCount` / `tasksRescheduledTitle` / `tasksMovedCount` keys.
- **Calendar-sync notification** (`App.jsx` `syncAll`): this message was assembled from `successes`/`errors` arrays, joined with `' and '`, and folded into three wrapping sentences.
  - Each clause (`"N events"`, `"N tasks"`, `calendar "name"`, `"calendar"`, `"task calendar"`) is now a translated fragment.
  - The join uses a new `localizedList()` in `localeFormatting.js` (`Intl.ListFormat`).
  - The three wrapping templates (`syncFailedWith` / `syncPartial` / `syncSuccess`) use a colon-plus-list shape (`"Synced: {{items}}"`), so no language has to make a list of fragments it didn't author agree in case or gender. A first draft that used the fragments as grammatical objects broke German dative and leaked `/a` and `(s)` artifacts into it/fr/pt.
  - Rendered in uk and pl, the bare-noun fragments need no inflection after the colon, because both languages take list items in the nominative there:
    ```
    uk  Синхронізовано: 1 подія — не вдалося: календар «Praca» і календар завдань
    pl  Zsynchronizowano: 1 wydarzenie — nie udało się: kalendarz „Praca” i kalendarz zadań
    ```
- **SpotlightModal**'s `"N results"` and **CalendarHeader**'s `"N deadlines"`.
- **`obsidianBinRestore.js` restore toast** (`binRestoreNoticeText`): both branches are now localized.
  - **Count 1** uses `sync.binRestoredOneDaily` (`{{title}}`, `{{date}}`) or `sync.binRestoredOneVault` (`{{title}}`). These are full sentences rather than a `{{where}}` fragment. The date goes through `formatLocalizedDate(…, { dateStyle: 'long' })`, parsed as local midnight.
  - **Count ≥ 2** uses `sync.binRestoredCount`. Count 1 is deliberately kept out of this plural family: Ukrainian's `one` category also covers 21, 31, … (`Intl.PluralRules('uk').select(21) === 'one'`), so `_one` must stay count-generic and never name a task.
  - `t` is an explicit parameter rather than a `useTranslation()` call, because `useObsidianSync.js`'s test files mock `'react'` down to `useEffect`/`useCallback`/`useRef`. The hook passes `i18n.t.bind(i18n)`, the same pattern as `useReminderEngine.js`.
  - **`binRestoreNoteLine`** (the text written into the user's Obsidian notes) is deliberately untouched. It is an idempotence key compared byte-for-byte across devices.
- **DeadlinePickerPopover**'s hardcoded `['Jan', 'Feb', ...]` month abbreviations are replaced with `formatLocalizedDate(viewDate, { month: 'short', year: 'numeric' })`.
- **uk/pl**: values for every key this PR adds, with `one/few/many/other` for all nine plural families.

## Merge order

This PR has no dependencies. #1781 has merged, and this branch is rebased on current `main` with its own uk/pl delta. It shares no conflicts with #1782 or #1783, so the three can land in any order.

## Test plan
- [x] `npx vitest run`: 5063 passing. The only failures are three in `solar.test.js`, which is timezone-dependent and fails on clean `main` too.
- [x] `src/locales.test.js`: coverage, the literal-`t()`-has-a-bundle-entry check, and placeholder parity across all 10 languages.
- [x] `localizedList` tests in `localeFormatting.test.js`: a single item, and two-item conjunctions in en/de/fr/zh-CN.
- [x] `binRestoreNoticeText` tests:
  - en at 1 (dated and undated), 2 and 5;
  - uk at 1 and 2;
  - uk at 21, asserting the count sentence and no title.
- [x] Mutation checks for the toast:
  - routing count 1 through the plural family fails 3 tests;
  - a title-bearing uk `_one` fails the 21 test;
  - `new Date('YYYY-MM-DD')` under `TZ=America/Los_Angeles` fails 2.
- [x] `useObsidianSync.binRestore.test.js` initialises i18next with the en bundle and asserts the translated single-task toast.
- [x] `eslint` on all touched files.
- [ ] No rendered component assertion for the `App.jsx` toasts, `SpotlightModal`, `CalendarHeader` or `DeadlinePickerPopover`. These are covered by the locale-coverage check, which the maintainer said is enough.


krelltunez:
This is the hardest section and it shows in the right places. Two things I want to call out before the findings, because they are the parts these PRs usually get wrong:

Leaving `binRestoreNoteLine` alone is exactly right. Its comment says locale-free on purpose because it is an idempotence key compared byte-for-byte across devices, and localizing it would have desynced every multi-device vault the moment two devices ran different languages. You read the comment and believed it.

The colon-plus-list reformulation of `syncFailedWith` / `syncPartial` / `syncSuccess` is the right instinct, and `localizedList` via `Intl.ListFormat` is better than the `join(' and ')` it replaces. Restructuring so no language has to agree with a fragment list it did not author is the same move that made your uk `monthlyOnOrdinalWeekday` work.

## One finding

`binRestoreNoticeText` is only half localized. The multi-task branch now goes through `t()`, but the single-task branch above it still returns hardcoded English:

```js
const where = r.dateStr ? `your ${r.dateStr} daily note` : 'your Obsidian vault';
return `Restored "${...}" from the recycle bin. Its line still exists in ${where}.`;
```

Rendered against a Ukrainian bundle, restoring one task versus two:

```
1 task : Restored "Завдання 1" from the recycle bin. Its line still exists in your 2026-09-22 daily note.
2 tasks: 2 tasks were restored from the recycle bin. ...
```

Once uk/pl have the new keys the second line becomes Ukrainian and the first stays English, which reads worse than uniform English would. The issue only named the plural string at line 86, so this is a fair reading of the checkbox rather than a miss, but the function is the unit a user experiences.

It has a concrete symptom: `sync.binRestoredCount_one` is unreachable. `binRestoreNoticeText` returns early at `length === 1`, so the count key is only ever called with 2 or more. You translated that `_one` form into eight languages and nothing can render it. Worth folding the single-task branch into the same key, which makes `_one` live and removes the split.

The embedded `{{dateStr}}` will need care when you do: a raw `2026-09-22` inside a sentence is its own locale problem, so `formatLocalizedDate` on the way in.

## Not blocking

`settings.importIcs` reuse: split it. You are right that de/fr/es/pt phrase it as an imperative for a menu item, and a toast title in the imperative reads like an instruction rather than a report. It is one extra key and it keeps the two registers independent.

`sync.calendarWord` / `taskCalendarWord` as bare-noun fragments: agreed, and agreed it is not yours to guess at. You will hit it when you translate these into uk/pl below, so if the colon-list shape turns out not to absorb it, say so and we will restructure that message rather than have you contort the nouns.

## What it needs before merge

#1781 merged, so the dependency chain in your description is stale: there is no "#1781 rebase" step coming. Each open PR now carries its own uk/pl delta instead, and I verified all three merge cleanly against each other and against current `main` with no git conflicts, so order does not matter.

For this PR specifically: rebase onto current `main` and add the uk/pl values for the ~30 new keys. Nine of them are plural families (`sync.icalImportedCount`, `icalRemovedCount`, `eventsCount`, `tasksCount`, `binRestoredCount`, `spotlight.resultsCount`, `app.tasksPlacedCount`, `tasksMovedCount`, `sched.deadlineCount`) and need `one/few/many/other` in both languages. Stacked with #1782 and #1783 on current `main` I get 5073 passing with only the uk/pl coverage failing, so that delta is the whole of what is left.

On your test-plan caveat about no rendered assertion for the toasts: that is honest and I am not going to require component tests for four call sites the locale-coverage check already guards. If you fold the single-task branch in, a direct unit test on `binRestoreNoticeText` at counts 1, 2 and 5 would be worth having, since that is where the split is.

polymatheiia:
Thanks, all addressed. Rebased onto current `main`.

## Bin-restore toast

The single-task branch now goes through `t()` as well. I did one thing differently from the suggestion: I didn't fold it into `binRestoredCount_one`. In Ukrainian the `one` category is n%10==1 && n%100!=11, so it covers 21, 31 and so on as well as 1:

```
new Intl.PluralRules('uk').select(21)  // 'one'
```

A `_one` that names `{{title}}` would therefore render one task's title as the toast for 21 restored tasks. That also means `binRestoredCount_one` was never fully dead. It is unreachable in en/de/fr/…, but live at 21/31/… in uk. So:

- Count 1 has two full-sentence keys of its own, `sync.binRestoredOneDaily` (`{{title}}`, `{{date}}`) and `sync.binRestoredOneVault` (`{{title}}`). They are two whole sentences rather than a `{{where}}` fragment, so no language has to decline "your daily note" / "your vault" into a slot it didn't author.
- `binRestoredCount` stays the ≥2 path, and every form is count-generic. The uk `_one` value is written for 21/31/… (plural "Їхні рядки…").
- `{{date}}` goes through `formatLocalizedDate(…, { dateStyle: 'long' })` from local midnight (`new Date(y, m-1, d)`, not the UTC parse of `'YYYY-MM-DD'`). Ukrainian long dates end in "р.", so the uk sentence is ordered to avoid a trailing "р..".

Tests on `binRestoreNoticeText`: en at 1 (dated and undated), 2 and 5; uk at 1 and 2; and uk at 21 asserting the count sentence and no title. Mutation-checked:
- routing count 1 back through the plural family fails 3;
- a title-bearing uk `_one` fails the 21 test;
- `new Date(dateStr)` under `TZ=America/Los_Angeles` fails 2.

`useObsidianSync.binRestore.test.js` now initialises the global i18next with the en bundle, since the single-task toast it asserts is translated now.

## `settings.importIcs` split

The new key `sync.icalImportTitle` is used for both toast titles (import and remove) in all 10 bundles, and `settings.importIcs` stays with the menu item.

## uk/pl

Values are in for all of the new keys, including one/few/many/other for all nine plural families. I rendered the composed calendar-sync messages to check `calendarWord` / `taskCalendarWord`:

```
uk  Синхронізовано: 1 подія — не вдалося: календар «Praca» і календар завдань
    Не вдалося синхронізувати: календар і календар завдань. …
pl  Zsynchronizowano: 1 wydarzenie — nie udało się: kalendarz „Praca” i kalendarz zadań
    Nie udało się zsynchronizować: kalendarz i kalendarz zadań. …
```

The colon-list shape absorbs it. After the colon, both languages take the items in the nominative, which is what the bare-noun keys already are, so no restructuring is needed.

Full suite 5063 passing. The only failures are the three in `solar.test.js`, which also fail on clean `main` here (timezone-dependent). Lint is clean.

krelltunez:
Merging. 5066/5066 against current `main`, lint clean.

You were right about `binRestoredCount_one` and I was wrong. Ukrainian's `one` category is 1, 21 and 31, so folding the single-task branch into it with a `{{title}}` would have rendered one task's title as the toast for 21 restored tasks. The key was never dead either, exactly as you say: unreachable in en/de/fr, live at 21/31 in uk. Two full-sentence keys for count 1 is the correct shape, and keeping every `binRestoredCount` form count-generic is what makes it hold.

I verified the rendering rather than taking the description on trust:

```
[uk] n=1  Завдання «Купити молоко» відновлено з кошика. Його рядок … за 23 вересня 2026 р. …
[uk] n=21 Відновлено з кошика 21 завдання. Їхні рядки і далі є у вашому сховищі. …
[pl] n=21 Przywrócono z kosza 21 zadań. …
[pl] n=22 Przywrócono z kosza 22 zadania. …
```

Agreement is right across the paradigm in both languages, and ordering the uk sentence so the long-date `р.` lands mid-sentence rather than against a full stop is a nice touch.

All three mutation claims reproduce here: routing count 1 back through the plural family fails 3, a title-bearing uk `_one` fails the 21 test, and `new Date(dateStr)` under `TZ=America/Los_Angeles` fails 2. I also confirmed the underlying bug is real rather than theoretical, since `new Date("2026-09-23")` formats as `22 вересня` in Los Angeles. Your suite passes under UTC, `America/Los_Angeles` and `Pacific/Kiritimati`.

The `sync.icalImportTitle` split and the composed calendar-sync check both look right, and your finding that the colon-list shape takes the nominative in both languages settles the `calendarWord` question without any restructuring.

That is sections 1 through 4 of #1763 done. Thank you for all of this, and for the issue that started it: the plural-category work in particular found a live Italian bug that predated any of these PRs.

PR #1788, opened 2026-09-23
feat(jobo): add the pure core model (slice 2)

Lcub3d:
Refs #1726 (slice 2), #1623, #1744, #1762.

This is the focused pure-core implementation for JOBO Slice 2.

Scope

This PR adds the pure "src/jobo/core.js" layer and tests.

It has no React, storage, sync wiring, task-store mutation, UI, or clock lookup.

Core owns:

- Do record construction and validation
- capture-once "title" and "planSnapshot"
- Timed / Untimed execution representation
- per-attempt progress and reassessment
- interval correction and tombstones
- ensure-present completion attempts
- deterministic "pickJoboRecord"
- Original Plan → Final Plan raw comparison
- Plan → Do comparison
- derived timing summaries

Do record contract

The record follows the merged #1744 ownership model: progress stays on each Do attempt.

Canonical progress values are:

- "started"
- "partial"
- "mostly"
- "completed"

"completeDoAttempt()" creates a "completed" attempt.

"reassessDoProgress()" reassesses an existing attempt to "started", "partial", or "mostly" without changing its interval or captured history.

"title" and "planSnapshot" are captured once and preserved. Later task renames, reschedules, persistence passes, remote applies, progress reassessments, or interval corrections never refresh historical snapshots.

Timed / Untimed Do

Do explicitly distinguishes:

- "timed" — a real positive execution interval
- "untimed" — execution is known to have happened, but duration was not measured

Untimed Do uses explicit null interval coordinates:

{
  timing: 'untimed',
  startTime: null,
  endDate: null,
  endTime: null,
}

Zero minutes is never used as an unknown-duration sentinel.

Untimed Do is still real execution evidence, so it suppresses "notStarted", but it does not participate in start/finish timing, duration comparison, or Allen interval relations.

The persistence design has been updated to include this discriminant.

Prototype import

"migrateLegacyDoRecord()" is an import boundary for prototype rows only.

Prototype progress values map as:

- "started → started"
- "partial → partial"
- "mostly → mostly"
- "complete → completed"

Prototype rows that predate the explicit Timed / Untimed discriminant are normalized before canonical validation.

Current canonical records do not pass through migration.

Comparison model

The canonical comparison keeps independent dimensions:

- Plan context: "planned / noPlan / unknown"
- Start: "early / onTime / late"
- Finish: "early / onTime / late"
- Duration: "shorter / onEstimate / longer"
- Execution: "singleSession / splitSessions"
- all 13 Allen interval relations

The compact product summary is multi-label:

- "withinPlan"
- "late"
- "longer"
- "split"
- "notStarted"
- "unplanned"

Valid combinations include:

- "late + longer"
- "late + split"
- "longer + split"
- "late + longer + split"
- "withinPlan + split"

"notStarted" is derived only when the current displayed Plan has fully elapsed and no live Do attempt exists.

"classifyAgainstPlan()" remains only as a compatibility adapter for the prototype vocabulary; the comparison rules live in "compareExecutionToPlan()".

Measured duration uses the union of overlapping Timed-Do intervals, so the same wall-clock minute is counted once.

Plan anchors

The analytical anchors are:

"Original Plan → Final Plan → Do"

"comparePlanAnchors(originalPlan, finalPlan)" returns:

- "startShiftMinutes"
- "finishShiftMinutes"
- "durationDifferenceMinutes"
- "durationRatio"

Provisional Plans remain in the existing task-owned dayGLANCE model:

- "deferrals" is the monotonic count
- "planTrail" is the bounded recent trail

Slice 2 does not add another revision counter or duplicate those task-level persistence / merge rules.

Merge contract

"pickJoboRecord()" remains schema-agnostic and deterministic:

"updatedAt → observedAt → canonical JSON"

Migration happens before merge.

Tests

Coverage includes:

- record validation
- capture-once snapshots
- per-attempt progress
- progress reassessment
- Timed / Untimed Do
- prototype migration
- deterministic merge
- tombstones
- repeated completion attempts
- Original Plan → Final Plan raw metrics
- Plan → Do timing comparison
- overlap / gap / split accounting
- Not Started
- tolerance policy
- multi-label summaries
- all 13 Allen interval relations

Current checks are green:

- Audit
- Lint
- Test
- Build
- iOS build

Out of scope

Still outside Slice 2:

- persistence and sync wiring
- backup / restore
- React / UI
- completion detection
- native task-completion integration
- JOBO checkbox behavior
- Carry Forward / successors
- mobile

The comparison surface is frozen here for later slices to consume.


Lcub3d:
One related note: I’ve completed a behavior-preserving structural split of the Slice 2 core and kept it isolated in a separate fork PR:

Lcub3d/dayGLANCE#14

The public API is unchanged, and CI / tests / build / iOS are all green.

I haven’t mixed the refactor into #1788 so the contract review can stay focused. Once the Slice 2 contract is settled, I can replay it onto the latest 'main' and upstream the refactor separately.

krelltunez (changes requested):
@Lcub3d thank you, this is a substantial piece of work and it is in the right shape: pure, tested, and it drops straight into #1762's controller, which I confirmed by running your pickJoboRecord, completeDoAttempt and tombstoneDoRecord through the real ledger. Timed versus untimed Do is a genuine improvement on the merged contract and I would like it written into the persistence doc as a follow-up. The subprocess timezone test is fine; that is the honest way to test it.

On completion ownership I would keep the merged contract, and the reason is structural rather than a preference. Moving completion to a Plan-level completionStatus needs a durable home, and the PR leaves that out of scope. On the task it is a new field, which is the resurrection hazard the ledger design exists to avoid, and it duplicates the binary completed flag with a conflict between them. In the ledger it is a second collection. Not persisted, the assessment is lost. The merged contract already gives you what you are protecting: an un-complete never touches the interval or the snapshot, it only changes the assessment on that attempt under its own updatedAt, which is what last-writer-wins is for. So please keep progress on the Do record, have completeDoAttempt write completed, add a reassess function for the other three values, and keep migrateLegacyDoRecord only for importing the prototype's JSON. As it stands a record in the merged shape fails validateDoRecord, so this is the one blocker.

Two smaller things: the new label constants are snake_case while the legacy ones and the codebase are camelCase, so camelCase throughout; and once ownership is settled, take the proposal framing out of both docs so they read as the contract.

The comparison model is more than slice 5 will use at first, which is fine because it is pure, but let us freeze that surface here rather than grow it. The structural split in your fork can follow as its own PR once this lands.

Lcub3d:
Addressed in "defd7c9".

- kept "progress" on each Do record
- "completeDoAttempt()" now creates "completed"
- added "reassessDoProgress()" for "started / partial / mostly"
- limited "migrateLegacyDoRecord()" to the prototype-import boundary
- changed the new multi-word serialized labels to camelCase
- removed the Plan-owned completion proposal from the core docs
- documented Timed / Untimed Do in the persistence contract
- froze the comparison surface here

I also rechecked the updated core against #1762's ledger path after these changes.

CI, tests, build, and iOS are all green.

krelltunez (approved):
Rechecked defd7c9 here: 210 tests across your files and #1762's pass together, lint is clean, and the revised core runs through the real ledger controller end to end, including commit, reassess to partial, a remote apply of the reassessed copy, a late duplicate observation losing to it, and the tie-break in both orders. Every point from the review is addressed, and folding the timing discriminant into the persistence doc in this PR was the better call. Approving. Please mark it ready and I will merge; the slice 3 sync wiring follows against this shape.

PR #1790, opened 2026-09-23
i18n: migrate remaining duration formatters to formatDuration(minutes, t)

polymatheiia:
Follow-up to #1783 for the rest of §3 of #1763. Moves the remaining hand-rolled "Xh Ym" formatters onto `formatDuration(minutes, t)`, so they read from the `common.duration*` keys in every language.

## Migrated
- `App.jsx`: focus-log tiles (`fmtMin`) and the glance-ahead `committedStr` (still empty at 0 minutes)
- `WeekView.jsx`: routine chips (`fmtDur` removed; `WeekViewColumn` gets its own `useTranslation()`)
- `GlanceSidebar.jsx`, `MobileGlanceSection.jsx`: per-frame available time
- `HyperGlanceBar.jsx`: elapsed time in the completion card
- `useNewTaskInput.js`: `%` duration suggestions, including the hardcoded English `Duration:` prefix (now `common.duration`)

## Behaviour changes
- WeekView routine chips render `1h 30m` rather than `1h30m`, matching the other sites.
- The `Duration:` separator is still a fixed `: `; languages that want a different colon style would need a follow-up.

## Not in scope
- `daySummary.js` `formatMinutes`: the separate pre-existing gap noted on #1783.
- `trmnl.js` (text for an external device) and `quickAddParser.js` (English display for an English-only parser).
- The `common.minutesShort` / `voice.minutesShort` "N min" sites, which are a different display ladder.
- `App.jsx` also has a local `formatDuration` built on `formatLocalizedDurationMinutes` (Intl, from #1782), which produces different output from `common.duration*` in some locales. I imported the new helper there as `formatTranslatedDuration` to avoid the clash. Which of the two should win is a decision I've left to you.

## Tests
- New test in `useNewTaskInput.locale.test.js` asserts the suggestions render through the translated ladder in uk (exact strings) and en. Reverting the hook change makes it fail.
- The other call sites are inline in components and have no test of their own.
- Full suite: 5090 passing; the only failures are the 3 in `solar.test.js`, which also fail on clean `main`. Lint clean. No locale keys were added or removed.

## Changes since last review
- `GlanceSidebar.jsx` and `MobileGlanceSection.jsx`: the glance-ahead `committedStr` now uses `formatDuration(minutes, t)` too, so the committed badge renders the same on every surface (it previously read e.g. `1h 30 Min.` vs `1 Std. 30 Min.` in German).
- `GlanceLocalization.test.jsx`: the zh-CN committed-duration assertion pinned the Intl output (`1小时30分钟`); it now expects the key-driven `1 小时 30 分钟`. Reverting either swap fails it (6 and 2 cases).
- Keeping the `formatTranslatedDuration` alias for now; migrating the remaining `formatLocalizedDurationMinutes` call sites and retiring the helper will be a separate PR.
- Full suite 5092/5095 (only the 3 pre-existing `solar.test.js` failures), lint clean.


krelltunez (commented):
5095/5095 against current `main`, lint clean, and I confirmed the only hand-rolled formatters left are the three you list as out of scope (`daySummary.js`, `trmnl.js`, `quickAddParser.js`). The scope notes match the tree exactly.

Flagging the `App.jsx` clash yourself, rather than quietly picking one, was the right call. Answering it below, but there is one thing to fix first that falls out of the same question.

## `committedStr` now renders two ways in adjacent surfaces

Three places compute `committedStr` from the same `glanceAhead.committedMinutes` and feed the same `glance.committedDuration` key. This PR migrates one of them:

```
src/App.jsx:7698                      formatTranslatedDuration  (this PR)
src/components/GlanceSidebar.jsx:1300       formatLocalizedDurationMinutes
src/components/MobileGlanceSection.jsx:1161 formatLocalizedDurationMinutes
```

The two ladders disagree, so the same badge reads differently depending on which surface drew it:

```
de   1 Std. 30 Min.   vs   1h 30 Min.
uk   1 г 30 хв        vs   1г 30х
```

Both of those files are already open in this PR for `availStr`, so it is the same one-line change in each, a couple of lines above what you already touched. Worth doing here rather than leaving a visible split behind.

## Which ladder should win

The translated `common.duration*` one, so migrate toward it rather than away.

The Intl path is not just a spacing difference. Rendering both across all ten locales:

```
uk    45 min   45 хв   vs   45х
de    60 min   1 Std.  vs   1h
es/fr/it       45 min  vs   45min
```

Ukrainian standard is `хв`; Intl's narrow form gives the bare `х`. German gets `Std.` from the keys against a mixed `1h 30 Min.` from Intl. The locale files are the place a native speaker can fix that; `Intl.NumberFormat`'s narrow units are not.

That said, it is not this PR's job to finish: `formatLocalizedDurationMinutes` still has ten call sites (`NotesSubtasksPanel`, `DayDial`, `TaskPlanHistory`, the gap markers in both glance surfaces, and App.jsx's local wrapper), and `localeFormatting.test.js` pins its current output as deliberate. So take the two `committedStr` lines here, and let the rest be another follow-up if you want it. Keeping the `formatTranslatedDuration` alias meanwhile is fine.

## Smaller

The `1h30m` to `1h 30m` change in WeekView routine chips is an improvement, not just a consistency fix.

Your note that no test covers the inline component call sites is fair, and I am not going to require one for a formatter swap the locale suite already guards. The `useNewTaskInput.locale.test.js` addition is the right place to have spent the effort, since that one carried the hardcoded English `Duration:`.

krelltunez:
Two clarifications on my last comment, before you start on either.

**I mischaracterised the German assertion.** I said `localeFormatting.test.js` pins the current output as deliberate. Reading it again, what was deliberate is the *separator*: the test is named for joining with a space "where the locale would list them with a comma or a word", and the helper's own comment says the narrow unit list yields a comma in German and the word "e" in European Portuguese, which read as lists rather than durations. Nobody chose `1h` over `1 Std.` for German. That is just what `unitDisplay: 'narrow'` returns.

So please do not treat `1h 30 Min.` as an output to preserve. The translated keys have no separator to fight in the first place, which makes that whole workaround unnecessary rather than something to port.

**The wider migration is a yes, and the scope is everything.** Migrate all ten remaining `formatLocalizedDurationMinutes` call sites onto `formatDuration(minutes, t)` and retire the helper, along with the assertions that exist only to pin its separator behaviour. One ladder, in the locale files, where a native speaker can fix it. The Ukrainian `45х` against the correct `45 хв` is the case that decides it: that is a correctness problem rather than a density preference, and it will recur in every language whose narrow units Intl guesses badly.

That also means the `formatTranslatedDuration` alias in `App.jsx` is temporary. Once the local wrapper at line 276 goes, the import can take its own name back.

Feasibility, since I checked before asking: all five files holding those call sites already call `useTranslation()`, so `t` is in scope or one hook away. A couple sit in subcomponents and may need their own hook, the way `WeekViewColumn` did in this PR.

Still a separate PR from this one. #1790 stands on its own and only wants the two `committedStr` lines.

polymatheiia:
Done: both committedStr sites in GlanceSidebar.jsx and MobileGlanceSection.jsx now use formatDuration(minutes, t), so all three surfaces agree. The zh-CN assertion in GlanceLocalization.test.jsx was pinning the Intl output (1小时30分钟); it now expects the key-driven 1 小时 30 分钟, and reverting either swap fails it (6 and 2 cases, respectively), so it guards both surfaces.

Thanks for the clarification on the German assertion. I'll do the full migration in a separate PR from fresh main: all ten call sites, retiring the helper and its separator assertions, and letting the App.jsx import take its own name back.

Issue #929, opened 2026-05-29
[Enhancement] i18n contributors wanted: help translate dayGLANCE into your language!


krelltunez:
## Summary
Add internationalization support to dayGLANCE so the UI can be translated into multiple languages. German is the first target locale based on user request and the strong German presence in the self-hosted community.

## Scope
- ✅ Integrate an i18n library (react-i18next)
- ✅ Extract all hardcoded UI strings into translation files
- ✅ Add browser locale detection with fallback to English
- ✅ Ship initial French translation
- ✅ Ship German translation
- ✅ Ship Chinese translation
- 🎯 Add more languages (contributors welcome!)

## Out of scope (for now)
- RTL language support
- Localized date/number formatting beyond what the library provides by default
- Translating marketing site / README

## Notes
- This is a solo side project so no committed timeline
- Open to contributors, especially native speakers willing to review translations

cedriclocqueneux:
Hello,
I stumbled upon this solution by chance, and I find it very interesting! I might be able to get started on the French translation. How should I go about it?

krelltunez:
**Update:** French is now merged (thank you, @cedriclocqueneux)! Looking for other contributors to add additional languages. **German is the next priority**, but all languages welcome.

If you'd like to contribute a translation, comment here to claim a language and let's expand the app together!

cedriclocqueneux:
I just sent you the German, Spanish, Italian, and Portuguese translations for the three solutions. Thanks, Claude ;-)

krelltunez:
This is amazing, thanks @cedriclocqueneux! I've pushed out new versions of lastGLANCE and lifeGLANCE. Now working on this one.

Lcub3d:
Hi! I prepared a Simplified Chinese (`zh-CN`) localization branch from the current `main`:

https://github.com/Lcub3d/dayGLANCE/tree/codex/zh-cn

It follows the existing `react-i18next` structure, keeps browser language detection and English fallback, and covers the main UI, settings, dialogs/toasts, onboarding, Electron application/tray menus, and Android native settings, widgets, notifications, and focus/share surfaces.

Per the concern raised in #948, Chinese date/time behavior is handled end to end rather than by translating date labels alone: Chinese date and weekday order, a Monday-first default, and a 24-hour default. `ai-prompts.js`, README/marketing copy, self-hosting files, and unrelated deployment changes are excluded from this branch.

[...]

PR #1792, opened 2026-09-23
i18n: retire formatLocalizedDurationMinutes in favour of formatDuration

polymatheiia:
Follow-up to #1790, as agreed in its review: migrates the remaining `formatLocalizedDurationMinutes` call sites onto `formatDuration(minutes, t)` and retires the helper, so every duration in the app reads from the `common.duration*` keys.

**Depends on #1790.** This branch is stacked on it, so until #1790 merges the diff also shows its three commits. Only the top commit (`3e67580f`) is new here. I'll rebase onto `main` once #1790 lands.

## Migrated
- `App.jsx`: the six stats tiles. The local `formatDuration` wrapper is removed, and the imported `formatDuration` takes its own name back (the `formatTranslatedDuration` alias is gone).
- `DayDial.jsx`: `formatMinutes` (block labels and the time left / starts in / ended ago strings).
- `GlanceSidebar.jsx`, `MobileGlanceSection.jsx`: the agenda gap markers (two per file).
- `TaskPlanHistory.jsx`: plan durations in the history panel.
- `NotesSubtasksPanel.jsx`: subtask durations.

All of these components already had `t` in scope; none needed a new hook.

## Removed
- `formatLocalizedDurationMinutes` from `localeFormatting.js`, with its separator workaround.
- The two tests in `localeFormatting.test.js` that pinned its output and separator behaviour.

## Behaviour changes
Every one of these sites now renders the translated ladder instead of Intl narrow units, e.g. German `1 Std. 30 Min.` instead of `1h 30 Min.`, Ukrainian `45 хв` instead of `45х`, Chinese `1 小时 30 分钟` instead of `1小时30分钟`. English is unchanged.

## Tests
- `DayDial.a11y.test.jsx`: the German block label now expects `1 Std.`.
- `GlanceLocalization.test.jsx`: the zh-CN gap marker now expects `1 小时 30 分钟`.
- New in `TaskPlanHistory.test.jsx`: renders the panel in Ukrainian and expects `1 г` / `15 хв`. The existing tests only checked English, which both formatters produce. Swapping an English-only formatter back in fails this test and nothing else.
- The `App.jsx` stats tiles and the `NotesSubtasksPanel` subtask duration have no test of their own; they are the same one-call swap.
- Full suite: 5091 passing; the only failures are the 3 in `solar.test.js`, which also fail on clean `main`. Lint clean. No locale keys added or removed.


krelltunez:
Merging. With #1790 in, this rebases down to two commits, merges clean, 5270/5270 against current `main`, lint clean.

The retirement is complete rather than mostly complete, which is the thing I actually checked: no reference to `formatLocalizedDurationMinutes` survives anywhere in `src/`, the `formatTranslatedDuration` alias is gone and the import has its own name back, and the only hand-rolled formatters left are the three you declared out of scope on #1790, `trmnl.js`, `quickAddParser.js` and `daySummary.js`.

Removing the separator workaround was the part worth proving rather than assuming, since that helper existed precisely to stop `Intl.ListFormat` rendering `1h, 30 Min.` in German and `1 h e 30 min` in European Portuguese. I rendered all ten locales and checked for a comma or a conjunction in any of them:

```
de      45 Min.   1 Std.   1 Std. 30 Min.   2 Std. 15 Min.
pt-PT   45 min    1 h      1 h 30 min       2 h 15 min
uk      45 хв     1 г      1 г 30 хв        2 г 15 хв
zh-CN   45 分钟    1 小时    1 小时 30 分钟     2 小时 15 分钟
```

None of them reintroduce it. The workaround became unnecessary rather than being dropped on faith, which is the right reason to delete a guard: hand-written keys have no list to format in the first place.

Adding the Ukrainian assertion to `TaskPlanHistory.test.jsx` was the right instinct too. Your note that the existing tests only checked English, which both formatters produce identically, is exactly why they would have passed through this change without noticing it.

German reading `1 Std. 30 Min.` instead of the mixed `1h 30 Min.`, and Ukrainian `45 хв` instead of `45х`, is now true everywhere rather than in the handful of places we had got to. That closes section 3 of #1763 properly, and with it every section of that issue except the quick-add keywords I declined.

Thank you for the whole run of this: two new languages, four sections of an issue you raised yourself, and a couple of bugs found along the way that nobody had noticed in any language.

krelltunez:
Correction to my last line: I said this closes section 3 and leaves only the quick-add keywords. Checking the issue's own file list against the merged tree, two items it names are still open.

`src/utils/daySummary.js` is listed in section 3 and still hand-rolls English `h`/`m`. It has more reach than its one function suggests: `TitlebarSummaryStrip`, `SummaryStrip` in seven places, and `App.jsx:7804-7807`, which precomputes those strings for the native side. The comment there says `formatMinutes` keeps the wording identical to the in-app strip, so the strip and the tray are wrong together and would be fixed together. You flagged it as out of scope on #1783 and again on #1790, both times correctly; it just never got picked back up.

`src/trmnl.js:196-197` is section 2's e-ink weekday and date, still `'en-US'`. I would leave this one. It is the case raised back in #929 about whether the device has CJK glyphs, so English there may be deliberate until someone can check on real hardware rather than something to fix blind.

So section 3 wants `daySummary.js` and nothing else, and I am leaving #1763 open for it. No rush, and no obligation on your side: you have done four sections and two languages already. If you would rather stop here, say so and I will take it.

The only wrinkle is that `daySummary.js` is a pure utility with no `t` of its own, so it needs threading in from its callers, the way `binRestoreNoticeText` did.

